PR #5, opened 2022-10-11
fix(chart): remove SC from template

HknLof:
Removing storage class from helm template
for postgres. This lets the underlying
distro use the actual default.


flippingbits (approved):
Great, thank you!

PR #9, opened 2022-10-17
Feat(pipeline): create initial pipeline implementation

ChrisRousey:
This PR is for the implementation of the Datacater Pipeline Concept.

- Messages are to be consumed from the definded stream-in
- Forwarded to a Datacater Python-runner
    * Here, it is processed (transformed and or filtered)
- The processed message is produced to the defined stream-out

The defined test will currently not work, as there is a bug in the currently deployed python-runner on the `/batch` endpoint. I fixed the endpoint in the PR. 

To test manually, the python runner and be run locally:
1. `pip3 install -r requirements.txt`
2. `python3 -m uvicorn runner:app`

And by then setting environment variables:
- `DATACATER_PYTHONRUNNER_PORT=8000`
- `DATACATER_PYTHONRUNNER_HOST=localhost`

### DIFF
--- a/pipeline/src/main/java/io/datacater/Pipeline.java
+++ b/pipeline/src/main/java/io/datacater/Pipeline.java
@@ -1,0 +1,93 @@
+package io.datacater;
+
+import io.datacater.exceptions.TransformationException;
+import io.smallrye.common.annotation.Blocking;
+import io.vertx.core.json.JsonObject;
+import io.smallrye.reactive.messaging.kafka.Record;
+
+import java.util.UUID;
+import javax.enterprise.context.ApplicationScoped;
+import javax.inject.Inject;
+
+import io.vertx.mutiny.core.Vertx;
+import io.vertx.mutiny.core.buffer.Buffer;
+import io.vertx.mutiny.ext.web.client.HttpRequest;
+import io.vertx.mutiny.ext.web.client.HttpResponse;
+import io.vertx.mutiny.ext.web.client.WebClient;
+import org.eclipse.microprofile.config.ConfigProvider;
+import org.eclipse.microprofile.reactive.messaging.Acknowledgment;
+import org.eclipse.microprofile.reactive.messaging.Incoming;
+import org.eclipse.microprofile.reactive.messaging.Outgoing;
+import org.jboss.logging.Logger;
+import org.jboss.resteasy.reactive.RestResponse;
+
+@ApplicationScoped
+public class Pipeline {
+  private static final Logger LOGGER = Logger.getLogger(Pipeline.class);
+
+  private final Integer DATACATER_PYTHONRUNNER_PORT =
+          ConfigProvider.getConfig()
+                  .getOptionalValue("datacater.python-runner.port", Integer.class)
+                  .orElse(50000);
+
+  private final String DATACATER_PYTHONRUNNER_HOST =
+          ConfigProvider.getConfig()
+                  .getOptionalValue("datacater.python-runner.host", String.class)
+                  .orElse("localhost");
+
+  private String host;
+  private Integer port;
+
+  private static final String PIPELINE_IN = "pipeline-in";
+  private static final String PIPELINE_OUT = "pipeline-out";
+  private static final String PIPELINE_ERROR_MSG = "Pipeline could not process message.\n Key: %s\n Value: %s";
+
+  WebClient client;
+
+  @Inject
+  void VertxResource(Vertx vertx) {
+    this.client = WebClient.create(vertx);
+  }
+
+  @Incoming(PIPELINE_IN)
+  @Outgoing(PIPELINE_OUT)
+  @Blocking
+  @Acknowledgment(Acknowledgment.Strategy.POST_PROCESSING)
+  public Record<UUID, JsonObject> processUUID(Record<UUID, JsonObject> message) {
+    // This is deliberately blocking
+    JsonObject transformedMessage = handleMessage(message);
+    return Record.of(message.key(), transformedMessage);
+  }
+
+  private JsonObject handleMessage(Record<UUID, JsonObject> message) {
+    HttpRequest<Buffer> request = client.post(getPort(), getHost(), "");
+    HttpResponse<Buffer> response = request.sendJson(message.value().encode()).await().indefinitely();
+
+    if(response.statusCode() != RestResponse.StatusCode.OK){
+      String errorMsg = String.format(PIPELINE_ERROR_MSG, message.key(), response.bodyAsJsonObject().encodePrettily());
+      LOGGER.error(errorMsg);
+      throw new TransformationException(errorMsg);
+    }
+
+    return response.bodyAsJsonObject().getJsonObject("value");
+  }
+
+  protected void setNetwork(int port, String host){
+    this.host = host;
+    this.port = port;
+  }
+
+  private String getHost(){
+    if(host != null){
+      return host;
+    }
+    return DATACATER_PYTHONRUNNER_HOST;
+  }
+
+  private int getPort(){
+    if (port != null){
+      return port;
+    }
+    return DATACATER_PYTHONRUNNER_PORT;
+  }
+}
--- a/pipeline/src/main/java/io/datacater/Pipeline.java
+++ b/pipeline/src/main/java/io/datacater/Pipeline.java
@@ -65,14 +65,13 @@ public class Pipeline {
             .putHeader("Content-Type", "application/json");
     HttpResponse<Buffer> response = request.sendJson(new JsonArray().add(getMessage(message))).await().indefinitely();
 
-    LOGGER.info(response.statusCode());
     if(response.statusCode() != RestResponse.StatusCode.OK){
-      String errorMsg = String.format(PIPELINE_ERROR_MSG, message.key(), response.bodyAsJsonObject().encodePrettily());
+      String errorMsg = String.format(PIPELINE_ERROR_MSG, message.key(), response.bodyAsJsonArray().encodePrettily());
       LOGGER.error(errorMsg);
       throw new TransformationException(errorMsg);
     }
 
-    return response.bodyAsJsonObject().getJsonObject("value");
+    return response.bodyAsJsonArray().getJsonObject(0).getJsonObject("value");
   }
 
   protected void setNetwork(int port, String host){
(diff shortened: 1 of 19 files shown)


pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,99 @@
...
+
+  private final String DATACATER_PYTHONRUNNER_HOST =
+          ConfigProvider.getConfig()
+                  .getOptionalValue("datacater.python-runner.host", String.class)
+                  .orElse("localhost");
+
+  private String host;
+  private Integer port;
+
+  private static final String PIPELINE_IN = "pipeline-in";
```
flippingbits, line 42: At the moment, we use `stream-in` in the pipeline spec. Shall we rename this here to stay consistent?

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,99 @@
...
+  private static final String PIPELINE_OUT = "pipeline-out";
```
flippingbits, line 43: At the moment, we use `stream-out` in the pipeline spec. Shall we rename this here to stay consistent?

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,99 @@
...
+  private static final String PIPELINE_ERROR_MSG = "Pipeline could not process message.\n Key: %s\n Value: %s";
```
flippingbits, line 44: Can we please log the metadata as well? For debugging purposes, it might be handy to know the offset/partition of the record.

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,99 @@
...
+    this.client = WebClient.create(vertx);
+  }
+
+  @Incoming(PIPELINE_IN)
+  @Outgoing(PIPELINE_OUT)
+  @Blocking
+  @Acknowledgment(Acknowledgment.Strategy.POST_PROCESSING)
+  public Record<UUID, JsonObject> processUUID(Record<UUID, JsonObject> message) {
+    // This is deliberately blocking
+    JsonObject transformedMessage = handleMessage(message);
```
flippingbits, line 59: At the moment, the pipeline processes record by record.

Could we turn this into micro-batching records for performance reasons? An user could still use a batch size of 1 if they would prefer the record-by-record processing.

flippingbits: If the record gets "filtered out", the Python runner returns `null`.
We should reflect this here.

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,99 @@
...
+    return Record.of(message.key(), transformedMessage);
+  }
+
+  private JsonObject handleMessage(Record<UUID, JsonObject> message) {
+    HttpRequest<Buffer> request = client.post(getPort(), getHost(), "/batch")
+            .putHeader("Content-Type", "application/json");
+    HttpResponse<Buffer> response = request.sendJson(new JsonArray().add(getMessage(message))).await().indefinitely();
+
+    if(response.statusCode() != RestResponse.StatusCode.OK){
```
flippingbits, line 68: We cannot rely on the status code of the response for detecting errors when working with the batch endpoint but need to investigate the `metadata`. If an error happens, it [will be included there](https://github.com/DataCater/datacater/blob/main/python-runner/runner.py#L188).

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,99 @@
...
+      String errorMsg = String.format(PIPELINE_ERROR_MSG, message.key(), response.bodyAsJsonArray().encodePrettily());
+      LOGGER.error(errorMsg);
+      throw new TransformationException(errorMsg);
+    }
```
flippingbits, line 72: Shall we log debug/info messages in case of no errors? Users might be interested in investigating the data that flows through a pipeline.

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,99 @@
...
+
+    return response.bodyAsJsonArray().getJsonObject(0).getJsonObject("value");
```
flippingbits, line 74: I am currently working record-level transforms which will allow users to transform the key as well.

Let's discuss how we can incorporate this here.

flippingbits (changes requested):
Thanks a lot for your great work! I am looking forward to using this essential feature.

I left a couple of comments which are many related to the [update of the pipeline spec](https://github.com/DataCater/datacater/tree/feature/pipeline-spec) that is happening in parallel.

pipeline/src/main/java/io/datacater/Pipeline.java
ChrisRousey: > At the moment, the pipeline processes record by record.
> 
> Could we turn this into micro-batching records for performance reasons? An user could still use a batch size of 1 if they would prefer the record-by-record processing.

I looked into this and wrote down some notes on how the processing would look to batch process a stream:

- They might lose their order going back and forth to python runner - still need to test this - probably need to manually enforce order based on offset
- also, if there is an exception for one record and we want to nack it, the entire batch is nacked.
- Depending on acknowledgment strategy (i.e. processing continues on nack), this could be fatal
- The processing from our side requires us to transform the messages to and from lists multiple times
    - we receive as a list<UUID, JsonObject>
    - we need to forward to python-runner as jsonarray<jsonobject> which requires us to iterate over list and transform inside pipeline
    - we receive jsonobject from python runner containing multiple nodes which each hold a message
    - we need to check the metadata of each message received to filter and log failed processing
    - we need to transform the jsonobject back into list<UUID, JsonObject> and probably sort by offset, so that 
    - the messages are back in the original order.

Do we think the python-runner performance boost is high enough to even out the extra processing done in pipeline?

flippingbits: > I looked into this and wrote down some notes on how the processing would look to batch process a stream:

Thanks a lot. That is very helpful! :)
 
>     * They might lose their order going back and forth to python runner - still need to test this - probably need to manually enforce order based on offset

I feel that this is fine because records end up in a different topic/partition, where we cannot guarantee any particular order.

>     * also, if there is an exception for one record and we want to nack it, the entire batch is backed.

True! That is a downside. We must ack all records and handle exceptions inside the pipeline, e.g., by logging the exception/error, by writing the affected records to a dead-letter queue, or any other mean. 

>     * Depending on acknowledgment strategy (i.e. processing continues on nack), this could be fatal

Agreed. We should start with a simple acknowledgment strategy and extend to other strategies in the future.

>     * The processing from our side requires us to transform the messages to and from lists multiple times
>       
>       * we receive as a list<UUID, JsonObject>
>       * we need to forward to python-runner as jsonarray which requires us to iterate over list and transform inside pipeline
>       * we receive jsonobject from python runner containing multiple nodes which each hold a message
>       * we need to check the metadata of each message received to filter and log failed processing
>       * we need to transform the jsonobject back into list<UUID, JsonObject> and probably sort by offset, so that
>       * the messages are back in the original order.
> 
> 
> Do we think the python-runner performance boost is high enough to even out the extra processing done in pipeline?

Yes, this should still give us a huge performance boost. 🚀  Also, from my gut feeling, handling one JSON array containing X JSON objects should not take much more time than handling X single JSON objects, if we are looking at throughput. We might want to test it though to have more confidence.

pipeline/src/main/java/io/datacater/Pipeline.java
ChrisRousey: i updated the logic, now our key is sent to python runner and the key sent back from python runner is used when forwarding to stream out. So this should be possible once we add key transformations to the python runner

pipeline/src/main/java/io/datacater/Pipeline.java
ChrisRousey: Do you mean log every messaged that passes through, or explicitly logging when there has been no error returned by python-runner batch?

Either way, i don't have a strong opinion to either options. We can log the extra information but since i don't have much practical data analysis experience, i don't know if this adds any extra value or not 🤷🏼

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: Great, thanks! :)
I already implemented them in 20e7ec2fff820214c4cdcf7a6310a470d3cf9caa, they are not yet part of the `main` branch though.

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: Let's leave it out for now. Users could inspect the sink stream to get insights into the processed data.

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: Yay! :) This will be helpful.

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: Re 1: Makes sense, thanks!
Re 2: I believe this is a bug, which has been fixed with c7cfbac1cfccd131d34a5fb93a18696dd7f40330 (that is not yet merged into main).

HknLof (approved):

HknLof:
Thanks for the feature :)

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -0,0 +1,114 @@
...
+  @Incoming(PipelineConfig.STREAM_IN)
+  @Blocking
+  @Acknowledgment(Acknowledgment.Strategy.POST_PROCESSING)
+  public void processUUID(ConsumerRecords<JsonObject, JsonObject> messages) {
+    // This is deliberately blocking
+    handleMessages(messages);
+  }
+
+  private void handleMessages(ConsumerRecords<JsonObject, JsonObject> messages) {
+    HttpRequest<Buffer> request = client.post(getPort(), getHost(), PipelineConfig.ENDPOINT)
```
flippingbits, line 50: One thing that I just noticed: Some Python transforms might need longer than expected, requiring us to abort the evaluation of the pipeline spec in the runner.

I believe we should add a timeout here.  A reasonable default might be something like 60 seconds, but we should definitely make it configurable.

ChrisRousey: added ✅

flippingbits: Great, thanks a lot. LGTM 💯

flippingbits (approved):
🍻

PR #13, opened 2022-10-21
Feat(deployment) - Create Deployment Endpoint and K8 Deployment Logic

ChrisRousey:
This PR is for the implementation of the Deployment Endpoint, which is to be used for creating Kubernetes Deployments of our `Datacater/pipeline` image


platform-api/src/main/java/io/datacater/core/deployment/DeploymentConfig.java
```diff
@@ -0,0 +1,9 @@
+package io.datacater.core.deployment;
+
+public record DeploymentConfig(
```
flippingbits, line 3: This looks like Streams-related configuration. Why is it replicated here for deployments?

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -1,49 +1,143 @@
...
+  }
+
   @GET
-  public Uni<List<DeploymentEntity>> getDeployments() {
-    return sf.withSession(
-        session ->
-            session.createQuery("from DeploymentEntity", DeploymentEntity.class).getResultList());
+  @Path("{deploymentName}/logs")
+  @Produces(MediaType.TEXT_PLAIN)
+  public Uni<String> getLogs(@PathParam("deploymentName") String deploymentName) {
```
flippingbits, line 39: Can we offer server-sent events for the logs as well? This would make the communication between clients (CLI or UI) and the backend much more efficient.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
```diff
@@ -5,18 +5,48 @@
...
 
-public record DeploymentSpec(Map<String, String> deployment) {
-  public static JsonNode serializeDeploymentSpec(Map<String, String> deployment)
-      throws JsonProcessingException {
+public class DeploymentSpec {
+  @JsonProperty("pipeline-in")
+  private final String pipelineIn;
+
+  @JsonProperty("pipeline-out")
+  private final String pipelineOut;
```
flippingbits, line 15: Would it make sense to rename `pipeline-{in,out}` to `stream-{in,out}`? The current name sounds like a deployment connects with two pipelines.

Also, we should reference the UUIDs of the streams.

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -0,0 +1,247 @@
...
+            .withRequests(StaticConfig.RESOURCE_REQUESTS)
+            .withLimits(StaticConfig.RESOURCE_LIMITS)
+            .endResources()
+            .withVolumeMounts(
+                new VolumeMountBuilder()
+                    .withName(volumeName)
+                    .withMountPath(StaticConfig.MOUNT_PATH)
+                    .build())
+            .endContainer()
+            .withVolumes(
```
flippingbits, line 101: What is the purpose of the volume?

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -0,0 +1,247 @@
...
+            .endMetadata()
+            .withNewSpec()
+            .addNewContainer()
+            .withName(name)
+            .withImage(StaticConfig.EnvironmentVariables.FULL_IMAGE_NAME)
+            .withImagePullPolicy(StaticConfig.PULL_POLICY)
+            .withEnv(variables)
+            .withNewResources()
+            .withRequests(StaticConfig.RESOURCE_REQUESTS)
+            .withLimits(StaticConfig.RESOURCE_LIMITS)
```
flippingbits, line 93: Can we make deployment configuration, like resource request/limits or number of replicas, configurable via the spec of the DataCater Deployment? This would add a lot of flexibility to deploying pipelines :)

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -0,0 +1,247 @@
...
+    if (exists(name)) {
+      return name;
+    }
+
+    List<EnvVar> variables = getEnvironmentVariables(dcDeployment, streamIn, streamOut);
+
+    var deployment =
+        new DeploymentBuilder()
+            .withNewMetadata()
+            .withName(name)
```
flippingbits, line 51: I am not sure whether we should let users name Kubernetes resources. Let me think about it for a bit :)

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -0,0 +1,247 @@
...
+      DatacaterDeployment dcDeployment,
+      PipelineEntity pe,
+      StreamEntity streamIn,
+      StreamEntity streamOut) {
+    final String name = getName(dcDeployment.name());
+    final String volumeName = name + StaticConfig.VOLUME_NAME_SUFFIX;
+    k8NameSpace.create();
+
+    if (!pe.equals(new PipelineEntity())) {
+      k8ConfigMap.getOrCreate(name, pe);
```
flippingbits, line 39: I am not sure whether we should let users name Kubernetes resources. Let me think about it for a bit :)

flippingbits (changes requested):
Thanks for the great work!

I left a couple of comments and change requests.

I still need to think a bit about whether we should let users name the Kubernetes resources, e.g., config maps, or deployments. I am leaning towards a "no" but will come back with more details soon.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentConfig.java
ChrisRousey: Thanks for pointing this out. I had made it to give the stream config to k8 class. Don't actually use it anymore, deleted.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
ChrisRousey: > Would it make sense to rename `pipeline-{in,out}` to `stream-{in,out}`? The current name sounds like a deployment connects with two pipelines.

Thanks for catching this, it actually needs to be stream instead of pipelin (the config at least, since thats the naming in pipeline image)

 
> Also, we should reference the UUIDs of the streams.

I guess we need to rethink this. Right now we're referencing the stream names and using it when setting up a deployment. But we don't actually need to reference either here? We reference stream-in/out UUID in pipeline and the streamNames are found in stream object. I guess it depends on where we want our source of truth, inside deployment or inside pipeline.

My opinion: We could remove deployment-spec all together. User only need to enter pipeline UUID when creating a deployment, from there we get all the info we need. The only thing we would need is a deployment name -> can use pipeline name

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
ChrisRousey: As far as I know, that's the standard when [mounting configMaps](https://kubernetes.io/docs/concepts/configuration/configmap/#using-configmaps).

Out manifest would look (mostly) like this:
```
apiVersion: apps/v1
kind: Deployment
metadata:
  annotations:
    deployment.kubernetes.io/revision: "1"
  creationTimestamp: "2022-09-30T17:19:19Z"
  generation: 1
  labels:
    datacater.io/app: datacater-pipeline
    datacater.io/pipeline: "1"
    datacater.io/revision: "1"
  name: string3
  namespace: datacater
  resourceVersion: "1029646"
  uid: c9c2b8b5-da41-4d66-930d-e867bda91963
spec:
  minReadySeconds: 2
  progressDeadlineSeconds: 600
  replicas: 1
  revisionHistoryLimit: 10
  selector:
    matchLabels:
      datacater.io/app: datacater-pipeline
      datacater.io/pipeline: "1"
      datacater.io/revision: "1"
  strategy:
    rollingUpdate:
      maxSurge: 25%
      maxUnavailable: 25%
    type: RollingUpdate
  template:
    metadata:
      creationTimestamp: null
      labels:
        datacater.io/app: datacater-pipeline
        datacater.io/pipeline: "1"
        datacater.io/revision: "1"
    spec:
      containers:
      - env:
        - name: MP_MESSAGING_OUTGOING_PIPELINE-OUT_TOPIC
          value: string
        - name: MP_MESSAGING_INCOMING_PIPELINE-IN_TOPIC
          value: string
        image: ghcr.io/datacater/pipeline:1
        imagePullPolicy: IfNotPresent
        name: string3
        resources:
          limits:
            memory: 4Gi
          requests:
            cpu: 100m
            memory: 1536Mi
        terminationMessagePath: /dev/termination-log
        terminationMessagePolicy: File
        volumeMounts:
        - mountPath: /usr/app
          name: string3volume
      dnsPolicy: ClusterFirst
      restartPolicy: Always
      schedulerName: default-scheduler
      securityContext: {}
      terminationGracePeriodSeconds: 30
      volumes:
      - configMap:
          defaultMode: 420
          name: string3
        name: string3volume
status:
  availableReplicas: 1
  conditions:
  - lastTransitionTime: "2022-09-30T17:19:22Z"
    lastUpdateTime: "2022-09-30T17:19:22Z"
    message: Deployment has minimum availability.
    reason: MinimumReplicasAvailable
    status: "True"
    type: Available
  - lastTransitionTime: "2022-09-30T17:19:19Z"
    lastUpdateTime: "2022-09-30T17:19:22Z"
    message: ReplicaSet "string3-7475c8dfb9" has successfully progressed.
    reason: NewReplicaSetAvailable
    status: "True"
    type: Progressing
  observedGeneration: 1
  readyReplicas: 1
  replicas: 1
  updatedReplicas: 1
```

platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
flippingbits: Yes, makes sense 👍 Let's remove this information from the deployment and use the config from the pipeline spec.

When passing the configuration to Kubernetes, would it make sense to "inline" the stream config in the config map?

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
flippingbits: Thanks for the clarification.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
ChrisRousey: > When passing the configuration to Kubernetes, would it make sense to "inline" the stream config in the config map?

you mean with configmap instead of passing environment variables?

flippingbits: Yeah, since we do not know all config options for sure, it might be more flexible. What do you think about it?

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
ChrisRousey: Sounds good. Could I open a ticket for this though? I think it could be a good first Ticket for our new Engineer, as it it touches on working with releavant subjects but doesn't require a deeper understanding of the tech behind it. And i guess it isn't a show stopper for our scope for the next three weeks

flippingbits: great idea!

platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
ChrisRousey: Smallrye sadly isn't that flexible as far as i know. It requires stream-in and stream-out config to be set upon startup.

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
ChrisRousey: created [a ticket](https://github.com/DataCater/datacater/issues/15) for this

flippingbits: Thank you very much!

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -1,49 +1,162 @@
...
+import javax.ws.rs.sse.OutboundSseEvent;
+import javax.ws.rs.sse.Sse;
+import javax.ws.rs.sse.SseEventSink;
 import org.hibernate.reactive.mutiny.Mutiny;
+import org.jboss.logging.Logger;
 
 @Path("/api/alpha/deployments")
 @RolesAllowed("dev")
 @Produces(MediaType.APPLICATION_JSON)
-@SecurityRequirement(name = "apiToken")
```
flippingbits, line 20: This annotation is needed for the token-based authentication to show up in the swagger documentation. Any reason why you removed it?

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -1,49 +1,162 @@
...
+@RequestScoped
 public class DeploymentEndpoint {
+  private static final Logger LOGGER = Logger.getLogger(DeploymentEndpoint.class);
   @Inject Mutiny.SessionFactory sf;
 
+  @Inject KubernetesClient client;
+
+  @GET
+  @Path("{deploymentName}")
+  public Uni<ObjectMeta> getStream(@PathParam("deploymentName") String deploymentName) {
```
flippingbits, line 39: Please rename the method to `getDeployment`.

Also, we should reference deployments by their UUID. Deployment names are not guaranteed to be unique, I guess?

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -1,49 +1,162 @@
...
+
+  @GET
+  @Path("{deploymentName}/logs")
+  @Produces(MediaType.TEXT_PLAIN)
+  public Uni<String> getLogs(@PathParam("deploymentName") String deploymentName) {
+    return Uni.createFrom().item(getDeploymentLogs(deploymentName));
+  }
+
+  @GET
+  @Path("{deploymentName}/watchLogs")
```
flippingbits, line 51: In the path, please use lower-cased words with a hyphen as separator. Here, it would boil down to `{deploymentName}/watch-logs`.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -1,49 +1,162 @@
...
   }
 
   @POST
-  @RequestBody
-  @Consumes(MediaType.APPLICATION_JSON)
-  public Uni<Deployment> createDeployment(Deployment deployment) throws JsonProcessingException {
-    DeploymentEntity de = new DeploymentEntity(deployment.spec());
-    return sf.withTransaction(
-        (session, transaction) -> session.merge(de).replaceWith(() -> deployment));
+  @Path("{pipelineUuid}")
```
flippingbits, line 78: The `pipelineUuid` should not be part of the path but be passed as, for instance, a URI parameter, e.g., `POST /deployments/?pipeline_uuid=$pipelineUuid`

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -1,49 +1,162 @@
...
-    Root<DeploymentEntity> root = delete.from(DeploymentEntity.class);
-    delete.where(cb.equal(root.get("spec"), spec));
+  @Path("{deploymentName}")
+  public Uni<Response> deleteDeployment(@PathParam("deploymentName") String deploymentName) {
+    return Uni.createFrom().item(deleteK8Deployment(deploymentName));
+  }
+
+  @PUT
+  @Path("{pipelineUuid}")
+  public Uni<String> updateDeployment(@PathParam("pipelineUuid") UUID pipelineId) {
```
flippingbits, line 91: At some point, one pipeline could have multiple deployments. When updating a deployment, we should reference it by the deployment's uuid.

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
flippingbits: I would suggest that we do not let users name Kubernetes deployments/resources to avoid unexpected conflicts.

flippingbits (changes requested):
Great work, thanks a lot! 💯 

I left a couple of comments on the endpoints for next week.

Have a great weekend! 🍻

HknLof (commented):
Couple of comments from my side:

1. If we propagate each resource / state to Kubernetes, then we need to have unique names or a second duplicated field under labels: `datacater.io/display-name: deploymentName`.

2. Every value, that is static, should be referenced and not build in-line. E.g. labels are build with `Map.of(...)`, this should be moved to static field or constant.

I will mark other spots, once you have paired up on this PR :).

Thank you both for your work!

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
ChrisRousey: datacater-deployment-UUID.Random()
Put name in a label

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
ChrisRousey: reference the new deployment uuid and change pipelineuuid from `@PathParam` to `@QueryParam`

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -77,24 +75,24 @@ public Uni<ListMeta> getDeployments() {
...
+  public Uni<Response> deleteDeployment(@PathParam("uuid") UUID deploymentId) {
+    return Uni.createFrom().item(deleteK8Deployment(deploymentId));
   }
 
   @PUT
-  @Path("{pipelineUuid}")
-  public Uni<String> updateDeployment(@PathParam("pipelineUuid") UUID pipelineId) {
+  @Path("{uuid}")
+  public Uni<UUID> updateDeployment(
+      @PathParam("uuid") UUID deploymentUuid, @QueryParam("pipelineUuid") UUID pipelineId) {
```
flippingbits, line 91: Since `pipelineUuid` is part of the deployment spec, could we also accept a deployment spec as request body here, assuming that we do this as well for the "update" endpoints of other resources?

ChrisRousey: We removed the deploymentSpec since we aren't persisting it and it only contained pipelineID and stream-in/stream-out which are already referenced in the PipelineObject. 

I'll re-add it to stay consistant with the other endpoints and since we will need it again at some point for K8 resource config and to change the deployments stream-in/out

platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
ChrisRousey: there seems to be two options for configuring our streams in deployment:

1. We map everything separately in the environment variables
2. We use [kafka-config](https://quarkus.io/guides/kafka#kafka-configuration-resolution)
    * We can either add one environment variable with a map, and these are mapped through `@Identifier("my-configuration")`-method
    * We add a configmap with map these with `@Identifier("my-configuration")`-method

ChrisRousey: I've updated deployment and pipeline to now forward the kafka-config to our pipeline.
`boostrap.servers` and `key/value.(de)serializer` are take from `StreamSpec`, but will be overwritten if these values are found in the `DeploymentSpec`


```
{
  "pipeline": "dc161a69-fa49-4b1a-b1b1-6d8246d50d72",
  "stream-in-config": {
    "bootstrap.servers": "localhost:9092",
    ....
  },
  "stream-out-config": {
    "bootstrap.servers": "localhost:9092",
    .....
  }
}
```
is the format we are looking for in deploymentSpec, where `stream-in-config` and `stream-out-config` are optional

flippingbits (changes requested):
I have a couple of small change requests. Feel free to have a look at them and see what makes sense to you :)

Feeling good about being close to the finish line for this PR 🏁

flippingbits (approved):
LGTM, Great Work! 💯 🥳

PR #16, opened 2022-10-28
Feature/pipeline spec

### LINKED ISSUE #1: Support record-level Python transforms
DataCater allows users to filter and transform data using Python functions. Despite having access to entire records, they can change only one column at a time. Having the possibility to change entire rows at a time is useful for many use cases, one of them being the flattening of JSON objects.

Our current format for transforms is as follows:

    def transform(row, value, config):
      return value

We should keep the same format but allow to return entire rows, e.g.:

    def transform(row, value, config):
      return row

### LINKED ISSUE #2: Let users change the names of transformation steps in the UI
Transformation steps consist of a name and a set of transforms. Naming transformation steps helps users to better identify them later on.

We should allow users to change the name of transformation steps through the UI. At the moment, this is only possible by working with our API directly.

### LINKED ISSUE #14: Proposal for a more flexible pipeline specification
At the moment, our pipeline spec is very opinionated and limited with regards to defining filters and transforms. It is not flexible enough to serve for general-purpose streaming data pipelines.

The main limitations of our current spec are:
* Pipelines can define a set of filters and a set of transforms. Transforms are always applied after filters. Users cannot chain transforms and filters in an arbitrary order.
* Filters and transforms only operate on the values of Kafka, not the keys.

The current pipeline spec looks as follows ([example](https://github.com/DataCater/datacater/blob/main/python-runner/example.pipeline.yaml)):

    ---
    spec      :
      filters: []
      transformationSteps:
      -   name: First
          transformations:
          -   attributeName: name
              transformation: trim
          -   attributeName: email
              transformation: trim
              filter: not-empty
          -   attributeName: age
              transformation: add-column
              transformationConfig:
                attributeName: age
                defaultValue: 42

I propose changing our pipeline spec as follows:

* Allow users to build a sequence of pipeline steps, which can filter and transform at the same time. I.e., filters can be defined at any stage of a pipeline.
* Allow users to apply transforms/filters to single fields of records (this is already possible atm) as well as entire records in one go (this is not yet possible). To this end, a step can be of `kind: Filter` or `kind: Record`.

Speaking YAML, the spec could look as follows:
    
    ---
    spec      :
      steps:
      - kind: Field
        fields:
          name:
            transform:
              key: trim
          email:
            transform:
              key: hash
              config:
                algorithm: sha1
            filter:
              key: not-empty
      - kind: Record
        transform: user-defined-record-transformation
        config:
          code: |
            def transform(record):
              return record

Filters and transforms are referenced by their key.
They can be configured with the object `config`. `Field`-level steps allow to apply filters/transforms to single fields of the record's value while `Record`-level steps allow to operate on entire records, consisting of a `key`, a `value`, and `metadata`.

flippingbits:
This PR introduces a new, more flexible specification for our pipelines.

Instead of fixing the order of filters/transforms and solely operating on single fields of the values of records, we can now chain filters and transforms in an arbitrary manner and apply filters/transforms to both single fields and entire records at a time.

This PR got huge and fixes multiple issues; my apologies - will try to avoid this in the future.

Fixes #14 
Fixes #1
Fixes #2


filters/user-defined-record-filter/spec.yml
```diff
@@ -0,0 +1,15 @@
+---
+kind       : Record
+name       : User-defined filter
+key        : user-defined-record-filter
+config     :
+-   name: code
+    label: Code
+    type: text
+description: Custom user-defined filter defined in Python
```
HknLof, line 9: This description matches `filters/user-defined-filter/spec.yml` exactly. Please make sure to distinguish consistently between record-level and field-level operations.

python-runner/test_runner.py
```diff
@@ -0,0 +1,353 @@
+# Run tests: $ python3 -m pytest
```
HknLof, line 1: +1 Thanks for adding this here

HknLof (approved):

python-runner/test_runner.py
flippingbits: ❤️

filters/user-defined-record-filter/spec.yml
flippingbits: Done: d29a1aa98adeb3be0e6ed567b1d4a7f22d4033f9

flippingbits: Thanks! :)

HknLof (approved):

PR #17, opened 2022-10-29
feat(python): Update Python to 3.11

flippingbits:
Update the Python version used for our Python runner to 3.11 to stay up-to-date with the latest developments and benefit from the performance gains.


HknLof (commented):
Is the `build.sh` file intended?

flippingbits:
Yes, I made it executable. We can revert it though if it is preferred.

HknLof (approved):

PR #19, opened 2022-11-03
Pooling implementation

HknLof:
- Use distributed maps for managing python-runner pods [at the time of writing without locks]
- Get pods only from distributed map
  - Reset queue if empty by fetching pods via labels
  - If queue has not put into map yet, then add first

Cleanup and some more tests outstanding.


platform-api/src/main/java/io/datacater/core/kubernetes/PythonRunnerPool.java
```diff
@@ -116,206 +94,100 @@ private String prependSlash(String path) {
...
-  private Boolean isStatefulSetActive() {
-    int statefulSetCount =
+  private Deque<NamedPod> podsFromKubernetes() {
+    List<Pod> pods =
         kubernetesClient
-            .apps()
-            .statefulSets()
+            .pods()
             .inNamespace(DataCaterK8sConfig.NAMESPACE)
             .withLabels(DataCaterK8sConfig.LABELS)
```
flippingbits, line 183: Would a request to a pod, which is in the `Terminating` or `Waiting` state, instantly fail or wait until the pod is `Running`?

If it would immediately fail, it could make sense to only select pods that are in a [Running](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-state-running) state.

flippingbits (changes requested):
Looks good to me, thank you very much for this important contribution!

I have one open question that I would like to discuss before merging this PR.

ChrisRousey:
I can't add it to my review, since it came from main, but `PythonRunnerPool` line 72 and 73 will throw en exception if there are no containers. Feel free to delete this comment if it isn't relevant because `pod.getSpec().getContainers()` wouldn't be empty

flippingbits (dismissed):
LGTM 🚀 Thanks for addressing my question.

platform-api/src/main/java/io/datacater/core/kubernetes/PythonRunnerPool.java
```diff
@@ -116,206 +94,101 @@ private String prependSlash(String path) {
...
+            .withField("status.phase", "Running")
             .list()
-            .getItems()
-            .size();
```
ChrisRousey, line 308: Could we come into the scenario, where we de-queue a runner which then executes for a while, reset the queue, and the runner get re-added and used, even though it is still executing?

I don't see anything marking the runners is in an `executing`-state.

If it's relevant, we could open an improvement issue.

ChrisRousey (commented):
Looks good to me, i just have one question for clarification

platform-api/src/main/java/io/datacater/core/pipeline/PipelineEndpoint.java
```diff
@@ -112,20 +114,24 @@ public Uni<Response> deletePipeline(@PathParam("uuid") UUID uuid) {
...
-              return Uni.createFrom().completionStage(previewSend);
-            })
-        .map(response -> response.body());
+        .onItem()
+        .transform(
+            Unchecked.function(
+                pod -> {
+                  HttpRequest preview = pod.buildPost(payload, "/preview");
+                  HttpResponse<String> previewSend =
+                      httpClient.send(preview, BodyHandlers.ofString());
```
flippingbits, line 129: We should introduce a (configurable) timeout here, [similar to the one by the pipeline](https://github.com/DataCater/datacater/blob/main/pipeline/src/main/java/io/datacater/Pipeline.java#L53).

flippingbits (changes requested):

platform-api/src/main/java/io/datacater/core/kubernetes/PythonRunnerPool.java
HknLof: We should test thoroughly as it might be already by way of ordering the `getMap` request by smallrye mutiny. Just in case I created a ticket https://github.com/DataCater/datacater/issues/27.

platform-api/src/main/java/io/datacater/core/pipeline/PipelineEndpoint.java
HknLof: Added without retry

platform-api/src/main/java/io/datacater/core/kubernetes/DataCaterK8sConfig.java
```diff
@@ -40,6 +40,10 @@ private DataCaterK8sConfig() {}
       ConfigProvider.getConfig()
           .getOptionalValue("datacater.pythonrunner.image.namespace", String.class)
           .orElse("datacater");
+  public static final long PYTHON_RUNNER_PREVIEW_TIMEOUT =
+      ConfigProvider.getConfig()
+          .getOptionalValue("datacater.pythonrunner.preview.timeout", Long.class)
+          .orElse(1000L);
```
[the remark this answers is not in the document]
HknLof: Yepp, fine by me. Will change asap to 10000L

platform-api/src/main/java/io/datacater/core/kubernetes/DataCaterK8sConfig.java
```diff
@@ -40,6 +40,10 @@ private DataCaterK8sConfig() {}
       ConfigProvider.getConfig()
           .getOptionalValue("datacater.pythonrunner.image.namespace", String.class)
           .orElse("datacater");
+  public static final long PYTHON_RUNNER_PREVIEW_TIMEOUT =
+      ConfigProvider.getConfig()
+          .getOptionalValue("datacater.pythonrunner.preview.timeout", Long.class)
```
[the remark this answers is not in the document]
flippingbits: Makes sense, also not a big fan of nesting :)

flippingbits (approved):
Thanks for implementing the changes! :) LGTM

PR #21, opened 2022-11-04
dynamic api path

ChrisRousey:
This PR updates the endpoint pathing to dynamically change when the version is updated.


flippingbits (approved):
Thanks for cleaning this up 👍  🤝

PR #33, opened 2022-11-10
Resets pooling correctly. 

HknLof:
Could not thoroughly test yet, once done will be converted to PR from draft.

### DIFF
--- a/load-image-k8s.bash
+++ b/load-image-k8s.bash
@@ -3,7 +3,7 @@
 set -eux
 
 NAME="datacater"
-VERSION="alpha.15"
+VERSION="local.3"
 START_DIR=$PWD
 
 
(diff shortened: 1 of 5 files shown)


HknLof:
This PR introduces:

- Better handling and async control flow for managing the python runner pool in a vertx cluster map.
- Introduce readiness and liveness probes for python runner instances created.
- Minor improvements

load-image-k8s.bash
```diff
@@ -3,20 +3,14 @@
 set -eux
 
 NAME="datacater"
-VERSION="alpha.15"
+VERSION="local.13"
```
flippingbits, line 6: Would it make sense to generate a unique image version, e.g., using the current timestamp, to avoid having to update this line with each change?

Consider it as food for thought, does not have to be implemented in this PR :)

flippingbits (approved):
LGTM, Thanks 👍

flippingbits (approved):

PR #34, opened 2022-11-10
feat(ui): Do not preview pipeline when inspecting source stream

flippingbits:
Avoid calling the `/pipelines/preview` endpoint when only inspecting the source stream.


HknLof (approved):

HknLof:
Thanks for decreasing amount of requests :)

PR #37, opened 2022-11-11
feat(ui): Improve the accessibility of Streams

flippingbits:
I improved the user experience and accessibility of the Streams form in the UI, making adding new or editing existing streams more straightforward. The changes incorporate feedback from a UX session with @ChrisRousey.

I recorded the usage of the new form, so you can get an impression of what it feels like:

[Demo Streams UI](https://www.loom.com/share/d67de71e573742bb82465eb17903f665)

Looking forward to your feedback! :)

### DIFF
--- a/ui/package.json
+++ b/ui/package.json
@@ -42,7 +42,7 @@
     "not ie <= 11",
     "not op_mini all"
   ],
-  "proxy": "http://127.0.0.1:8080",
+  "proxy": "http://127.0.0.1:8081",
   "engines": {
     "node": ">=12.22.0 <13",
     "npm": "~6.14.12"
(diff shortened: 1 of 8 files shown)


ChrisRousey (approved):
lets great!

I do want to mention though, that we also have the possability to enter an avro schema directly, without actually using a schema repo. I personally think this is only relevant for developers in a test environment, which would probably not primaraly use the UI, so i don't think it's necessary to implement the functionality here.

flippingbits:
Thanks a lot for your feedback (here and in the Zoom call)! :)

Although I have seen the use of Avro with a schema registry much more often than an "inline" Avro schema, we must still support inline schemas.

For the time being, users can provide it as a configuration under "Connection", using the consumer/producer API of Kafka. If we decide to have a "stream wizard", we should let users decide between using Avro with or without a schema registry and can then show the appropriate config option.

PR #43, opened 2022-11-16
Minor quality upgrade

ChrisRousey:
This PR is to:

- [x] remove some code smells
- [x] up the test coverage
- [x] add a 404 response to deployment delete endpoint, when deployment doesn't exist
- [x] pipeline doesn't need a stream-out/in. If doesn't exist in piepline, look in deployment spec. If none, throw exception.
- [x] add deployment name to object and add as label to deployment
- [x] fix error when creating an empty deployment


platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
```diff
@@ -1,20 +1,23 @@
...
 import org.eclipse.microprofile.openapi.annotations.media.Schema;
 
 public record DeploymentSpec(
+    @JsonProperty(value = "name") @JsonSetter(nulls = Nulls.AS_EMPTY) String name,
     @Schema(
             example =
                 """
             {
                 "pipeline": "dc161a69-fa49-4b1a-b1b1-6d8246d50d72",
                 "stream-in-config": {
```
flippingbits, line 16: I believe we should align the naming with `pipeline`.

So either do

```
{
  "pipeline": "uuid...",
  "stream-in": "uuid...",
  "stream-out": "uuid..."
}
```

or


```
{
  "pipeline-config": {},
  "stream-in-config": {},
  "stream-out-config": {}
}
```

platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
```diff
@@ -24,15 +24,15 @@
  *
  * <p>Optional connection options: - value.deserializer.schema - schema.registry.url
  *
- * <p>Optional config options: - auth... (TODO) - key.format (default: json) - num.partitions
- * (default: 3) - replication.factor (default: 1) - value.format (default: json)
+ * <p>Optional config options: - key.format (default: json) - num.partitions (default: 3) -
```
flippingbits, line 27: We need to update the JavaDoc here (rename `num.partitions` to `partition.count`.

flippingbits (changes requested):
LGTM except two minor issues 🍰  Would be great if you could fix them before we merge the PR.

flippingbits (approved):

Issue #47, opened 2022-11-17
Increase robustness of /api/alpha/deployments/{uuid}/watch-logs endpoint


flippingbits:
When working with the `/api/alpha/deployments/{uuid}/watch-logs` endpoint, I ran into timeouts very often.

We should investigate the behavior and try to make the endpoint more robust, enabling the efficient streaming of deployment logs to clients, like our CLI and UI.

HknLof:
I will try to replicate this problem.

flippingbits:
@HknLof Did you have a chance to replicate it? If not, I would suggest closing this stale issue.

Issue #46, opened 2022-11-17
Restrict number of lines returned by /api/alpha/deployments/{uuid}/logs


flippingbits:
It seems that the `/api/alpha/deployments/{uuid}/logs` endpoint returns all logs of the deployment. We should restrict it to the last X lines to prevent performance issues in both the server and the client.

We should make X configurable both via a query parameter (e.g., `/api/alpha/deployments/{uuid}/logs?limit=X`) and a configuration property in the [application.yml](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/resources/application.yml).

Issue #45, opened 2022-11-17
Let /api/alpha/deployments/{uuid}/logs return valid JSON


flippingbits:
At the moment, the endpoint `/api/alpha/deployments/{uuid}/logs` returns ND-JSON, which does not match [the returned media type](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java#L55) and must be manually parsed by a client, e.g.:

```
{"timestamp":"2022-11-17T05:28:09.485Z","sequence":5980,"loggerClassName":"org.apache.kafka.common.utils.LogContext$LocationAwareKafkaLogger","loggerName":"org.apache.kafka.clients.NetworkClient","level":"WARN","message":"[Consumer clientId=kafka-consumer-stream-in, groupId=f5c48013-3ff3-4faa-961f-f5ecf5bd215d] Connection to node -1 (localhost/127.0.0.1:9092) could not be established. Broker may not be available.","threadName":"smallrye-kafka-consumer-thread-0","threadId":19,"mdc":{},"ndc":"","hostName":"datacater-deployment-0a37998b-9f49-4bc1-9c4c-72d23b03aabf-2kqsd","processName":"quarkus-run.jar","processId":1}
{"timestamp":"2022-11-17T05:28:09.485Z","sequence":5981,"loggerClassName":"org.apache.kafka.common.utils.LogContext$LocationAwareKafkaLogger","loggerName":"org.apache.kafka.clients.NetworkClient","level":"WARN","message":"[Consumer clientId=kafka-consumer-stream-in, groupId=f5c48013-3ff3-4faa-961f-f5ecf5bd215d] Bootstrap broker localhost:9092 (id: -1 rack: null) disconnected","threadName":"smallrye-kafka-consumer-thread-0","threadId":19,"mdc":{},"ndc":"","hostName":"datacater-deployment-0a37998b-9f49-4bc1-9c4c-72d23b03aabf-2kqsd","processName":"quarkus-run.jar","processId":1}
[...]

PR #48, opened 2022-11-17
Implement the UI for Deployments

flippingbits:
This PR implements the UI for deployments. Deployments are now a top-level item in our navigation.

In the UI, you can:
* Create new deployments
* Update deployments
* Get deployments
* Delete deployments
* Read the logs of the deployments

## Try it out

You can explore and try out the changes as follows:
* Check out the PR branch
* Install the latest release into your default namespace:
``` bash
kubectl apply -f k8s-manifests/minikube-with-postgres-ns-default.yaml
```
* Add a port forward:
```bash
kubectl port-forward ui 8080:8080
```
* Spin up the ReactJS application:
```bash
cd ui
npm start
```
* Navigate to `http://localhost:3000` in your browser

## Todos

I left a couple of TODOs, which should be addressed in the future:
* #45
At the moment, the `/api/alpha/deployments/{uuid}/logs` endpoint returns ND-JSON, which needs to be [manually parsed to turn it into valid JSON](https://github.com/DataCater/datacater/blob/a1bbf4fbe630cb7935d482b2d7f34e6db0b3b9de/ui/src/reducers/deployments.js#L98). I suggest to let the endpoint return a valid JSON array.
* #46
At the moment, it seems that the `/api/alpha/deployments/{uuid}/logs` endpoint returns all logs of the deployment. We should restrict it to the last X lines to prevent performance issues in both the server and the client. We should make X configurable both via a query parameter (e.g., `/api/alpha/deployments/{uuid}/logs?limit=X`) and via a configuration property in the `application.yml`.
* #47
It seems that the SSE-endpoint, `/api/alpha/deployments/{uuid}/watch-logs`, is slightly unstable. Before using that in the UI, we should further investigate it and make it more robust.


ChrisRousey (approved):
🔥

PR #52, opened 2022-11-21
feat(ui): Improve API Call Widget

flippingbits:
This PR applies the following changes to our UI:

* Remove code duplicates and move the Header/API Call Widget to a separate ReactJS component, called `Header`, which is now used in most of our container components.
* Allow users to dynamically insert their authentication token into the generated `curl` commands, allowing them to copy the `curl` command and instantly use it. The token is already stored in the local storage of the browser and is valid for only a restricted time period, thus not creating an additional attack surface.


ChrisRousey (approved):
🔥 very cool and helpful feature

PR #54, opened 2022-11-21
feat(ui): Indicate preview errors

flippingbits:
When previewing a pipeline in the UI fails, let the user know and display an error message.

### DIFF
--- a/ui/src/containers/pipelines/EditPipeline.js
+++ b/ui/src/containers/pipelines/EditPipeline.js
@@ -14,6 +14,7 @@ import {
   Search,
   Table,
   Type,
+  X,
 } from "react-feather";
 import { Modal } from "react-bootstrap";
 import BaseTable, { AutoResizer } from "react-base-table";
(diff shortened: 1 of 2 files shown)


ChrisRousey:
@flippingbits 
What type of errors should this be catching?
I tested a pipeline, which has a stream-in only containing invalid JSON Objects `'{"name":"John", "age":30, "car":null}'`.
This throws an exception, but the frontend gets stuck in `Loading sample records...` state

flippingbits:
This catches errors that happen when previewing a pipeline, for instance, because the request to the python runner has not been successful.

You can test it by setting the number of replicas of the python runner pool to 1 and performing multiple preview requests, e.g., by switching between different steps, within a few seconds.

Good catch, I will add error reporting for your case as well! 🤝

flippingbits:
@ChrisRousey May I ask you to test your case again?

flippingbits:
Thanks a lot for the help and feedback :)

PR #55, opened 2022-11-21
fix(ui): Fix header height of table

flippingbits:
Fix the height of the table header for pipeline steps of kind "Field".

Before:
Screenshot 2022-11-20 at 16 43 11

After:
Screenshot 2022-11-20 at 16 42 41


ChrisRousey (approved):
looks good 👍🏼

PR #63, opened 2022-11-21
feat(deployment): add consumer group to stream-in

ChrisRousey:
This PR adds a consumer `group.id` to our pipelines `stream-in` when creating a deployment.
The `group.id` is the uuid of the deployment.
If users add a `group-id` to the stream-config, it will be overwritten by us.


flippingbits (approved):
LGTM 🔥 💯

flippingbits (approved):

PR #68, opened 2022-11-23
feat(pipeline): Handle ConnectExceptions

flippingbits:
DataCater deployments use a sidecar, `python-runner`, for evaluating pipeline specifications.

If the sidecar is not (yet) available, accessing it will throw a `java.net.ConnectException`, nacking the processed record.

This can be easily reproduced when using a `stream-in` with the consumer configuration `auto.offset.reset = earliest`. The sidecar takes longer to spin up than the Quarkus application, thus always leading to nacked messages.

This commit introduces two configuration options:
* `PipelineConfig.CONNECTION_RETRIES`: Number of retries to perform when accessing the sidecar throws a `ConnectException` (default: `10`)
* `PipelineConfig.CONNECTION_RETRY_WAIT`: Number of milliseconds to wait between different connection attempts, leaving time for the sidecar to become available (default: `1000`)

The Quarkus application now tries to perform up to `PipelineConfig.CONNECTION_RETRIES` connection attempts when accessing the sidecar before it fails and nacks the messages.

### DIFF
--- a/pipeline/src/main/java/io/datacater/Pipeline.java
+++ b/pipeline/src/main/java/io/datacater/Pipeline.java
@@ -10,6 +10,7 @@ import java.time.Duration;
 import java.util.UUID;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
+import java.net.ConnectException;
 
 import io.vertx.mutiny.core.Vertx;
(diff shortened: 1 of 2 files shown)


flippingbits:
Thanks for your help! :)

Issue #57, opened 2022-11-21
Add endpoint /deployments/{uuid}/metrics


flippingbits:
Introduce the API endpoint `/deployments/{uuid}/metrics` that shall return important metrics of the deployment in a [Prometheus-compatible format](https://prometheus.io/docs/instrumenting/writing_exporters/). We might use Smallrye's support for MicroProfile Metrics.

Exemplary metrics are:

* Throughput in records/s
* Throughput in bytes/s
* Latency

Issue #56, opened 2022-11-21
Add endpoint /deployments/{uuid}/health


flippingbits:
Introduce the API endpoint `/deployments/{uuid}/health` that shall return the health status of a given deployment.

PR #69, opened 2022-11-24
Feature(Deployment) - add metrics and health endpoints

### LINKED ISSUE #66: Clean up config map when deleting deployment
When deleting a Deployment from Kubernetes, we should also remove the associated Config Map.

ChrisRousey:
this PR is for added the health and metrics endpoints, as defined in #56 and #57 

It also fixes #66 by deleting the respective configmap when a deployment is deleted

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
+++ b/platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
@@ -11,7 +11,12 @@ import io.fabric8.kubernetes.client.KubernetesClient;
 import io.fabric8.kubernetes.client.dsl.LogWatch;
 import io.fabric8.kubernetes.client.dsl.RollableScalableResource;
 import io.smallrye.mutiny.Uni;
+import io.smallrye.mutiny.unchecked.Unchecked;
 import java.io.*;
+import java.net.URI;
+import java.net.http.HttpClient;
+import java.net.http.HttpRequest;
+import java.net.http.HttpResponse;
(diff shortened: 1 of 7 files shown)


platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -237,6 +290,21 @@ private String getDeploymentLogs(UUID deploymentId) {
     return k8Deployment.getLogs(deploymentId);
   }
 
+  private HttpRequest buildRequest(UUID deploymentId, String path) {
```
flippingbits, line 293: Please rename to something more specific, e.g.: `buildRequestForDeployment`.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -237,6 +290,21 @@ private String getDeploymentLogs(UUID deploymentId) {
...
+    K8Deployment k8Deployment = new K8Deployment(client);
+    String clusterIp = k8Deployment.getClusterIp(deploymentId);
+    String uriReady =
+        String.format(
+            "http://%s:%d%s",
```
flippingbits, line 298: We might want to make the protocol configurable, just in case.

flippingbits (changes requested):
LGTM! :) I have two very small remarks

flippingbits (approved):
Thanks for your great work. Developers will love these new endpoints :)

PR #75, opened 2022-11-28
Add health and metrics of deployments to UI

flippingbits:
This PR makes the information from the new `/api/alpha/deployments/{uuid}/health` and `/api/alpha/deployments/{uuid}/metrics` endpoints available in the UI under `/deployments/{uuid}`.


ChrisRousey (approved):
Looks great 🙏🏼 

Some Metrics wouldn't get caught sometimes, this is more a backend problem though i guess.

PR #77, opened 2022-11-28
Feat(rls): enable rls and tenantawareness for our postges resources

ChrisRousey:


ChrisRousey:
@HknLof @flippingbits 
What we are doing is intercepting the hibernate reactive calls and wrapping them inside another transaction after we set the tenantId. 
We are currently only setting the tenantId to DataCater.
I'm not 100% sure the wrapper will work with the current implementation since we are wrapping a 
`Mutiny.Transaction` inside another `Mutiny.Transaction`.

Nevertheless, hibernate reactive doesn't seem to play well with RLS in general. The code works as i would expect without having RLS activated in Postgres. Once we activate RLS, queries run endlessly without exception and never commit to the database. The same thing happens if i take the code from main, add a `tenant_id` field to out entities manually and enable  RLS from there.

ChrisRousey:
Code to enable RLS in Postgres:
```
ALTER TABLE public.deployments ENABLE ROW LEVEL SECURITY;
CREATE POLICY deployments_isolation_policy ON public.deployments
    USING (tenant_id = current_setting('app.current_tenant')::VARCHAR);
    
ALTER TABLE public.pipeline ENABLE ROW LEVEL SECURITY;
CREATE POLICY pipeline_isolation_policy ON public.pipeline
    USING (tenant_id = current_setting('app.current_tenant')::VARCHAR);
    
ALTER TABLE public.streams ENABLE ROW LEVEL SECURITY;
CREATE POLICY streams_isolation_policy ON public.streams
    USING (tenant_id = current_setting('app.current_tenant')::VARCHAR);
```

HknLof:
A small experiment running in postgres only. I let the pipeline table being created by our application and added a column `tenant_id` of type `text`.

From there, I used the following commands to create and enforce a policy. After that insert data and retrieve.

```sql
-- enforce row level security define policy to use the current setting 'app.current_tenant' for enforcement
ALTER TABLE public.pipeline FORCE ROW LEVEL SECURITY;
CREATE POLICY pipeline_isolation_policy ON public.pipeline
    USING (tenant_id = current_setting('app.current_tenant'));

SET app.current_tenant = 'cater';
SHOW app.current_tenant;

-- #=> cater

INSERT INTO public.pipeline (id, createdat, metadata, name, spec, updatedat, tenant_id)
VALUES ('ec146d8c-5111-4b34-83da-fd1799859a77', '2022-11-30 14:22:27.789001', '"{}"', 'cater',
        '"{\"steps\":[]}"', '2022-11-30 14:22:27.789002', 'cater');

SELECT * FROM public.pipeline;
-- #=> results in above row returned

SET app.current_tenant = 'data';
SHOW app.current_tenant;
-- #=> data

SELECT * FROM public.pipeline;
-- #=> results in above row returned NOT WANTED

```
Another oddity:

```sql
SELECT * FROM public.pipeline WHERE (tenant_id = current_setting('app.current_tenant'));
```

I expect the above sql query to return the wanted entry, but it does not return anything.

Do we need a second table? See postgres documentation
https://www.postgresql.org/docs/current/ddl-rowsecurity.html

@flippingbits any ideas how RLS is meant to work within postgres?

HknLof:
After creating a postgres role with the login privilege, I get the following result:

```sql
-- Create Role
CREATE ROLE rls
WITH PASSWORD 'rls' LOGIN;

-- we can skip this, if it works as intended
GRANT SELECT ON public.pipeline TO rls;
```

```sql
set app.current_tenant = 'cater';
select * from pipeline;
-- returns 1 row ✅ 
set app.current_tenant = 'date';
select * from pipeline;
-- returns 0 rows ✅  
```

We need to set `app.current_tenant` [also rename to `datacater.tenant`] before the initial query for our user, otherwise we are running into following error

```
postgres.public> select * from pipeline
[2022-12-01 14:26:30] [42704] ERROR: unrecognized configuration parameter "app.current_tenant"
```

HknLof:
Addressed by https://github.com/DataCater/datacater/pull/80

PR #78, opened 2022-12-02
Feature/introduce config resource

olis1996:
@HknLof @ChrisRousey 
I created an initial version of the config endpoint and started implementing some tests. However, I have a few questions regarding the implementation:

1.) Should i log if a config is created, deleted, updated etc?
2.) Do we accept the creation of a config if another one with the exact same properties already exists? (same name, kind, config)
3.) If the user tries to update/delete a not-existing config the API should return 404, right?
4.) In "ConfigEndpointTest.java, why is my setUp method executed after my endpoint test methods, although it is annotated with "@Before"?

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
+++ b/platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
@@ -49,7 +49,8 @@ public class ConfigEndpoint {
   @RequestBody
   @Consumes(MediaType.APPLICATION_JSON)
   public Uni<ConfigEntity> createConfig(Config config) throws JsonProcessingException {
-    ConfigEntity configEntity = ConfigEntity.from(config.name(), config.kind(), config.spec());
+    ConfigEntity configEntity =
+        ConfigEntity.from(config.name(), config.kind(), config.metadata(), config.spec());
 
     return sf.withTransaction((session, transaction) -> session.persist(configEntity))
         .replaceWith(configEntity);
--- a/platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
+++ b/platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
@@ -1,0 +1,82 @@
+package io.datacater.core.config;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import io.datacater.core.exceptions.ConfigNotFoundException;
+import io.smallrye.mutiny.Uni;
+import java.util.List;
+import java.util.UUID;
+import javax.inject.Inject;
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
+import org.hibernate.reactive.mutiny.Mutiny.SessionFactory;
+
+@Path("/config")
+@Produces(MediaType.APPLICATION_JSON)
+public class ConfigEndpoint {
+  @Inject SessionFactory sf;
+
+  @GET
+  public Uni<List<ConfigEntity>> getAllConfigs() {
+    return sf.withTransaction(
+        session ->
+            session
+                .createQuery("from ConfigEntity", ConfigEntity.class)
+                .getResultList()
+                .onItem()
+                .ifNull()
+                .continueWith(List.of()));
+  }
+
+  @GET
+  @Path("{uuid}")
+  public Uni<ConfigEntity> getConfig(@PathParam("uuid") UUID uuid) {
+    return sf.withTransaction(
+        session ->
+            session
+                .find(ConfigEntity.class, uuid)
+                .onItem()
+                .ifNull()
+                .failWith(
+                    new ConfigNotFoundException(
+                        String.format("No config found for uuid %s", uuid.toString()))));
+  }
+
+  @POST
+  @RequestBody
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Uni<ConfigEntity> createConfig(Config config) throws JsonProcessingException {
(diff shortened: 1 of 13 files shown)


ChrisRousey:
> 3.) If the user tries to update/delete a not-existing config the API should return 404, right?

Yeah, 404. Your ExceptionMapper looks correct.

> 4.) In "ConfigEndpointTest.java, why is my setUp method executed after my endpoint test methods, although it is annotated with "https://github.com/before"?

You are using Junit.Jupiter (which is part of junit5) to annotate your tests, but the @Before Annotation is for JUnit4.
You should use the JUnit5 Annotation `@BeforeAll` and can remove the `import org.junit.Before;` in line 17

platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+  @RequestBody
+  @Consumes(MediaType.APPLICATION_JSON)
+  public Uni<ConfigEntity> updateConfig(@PathParam("uuid") UUID uuid, Config config) {
+    return sf.withTransaction(
+        ((session, transaction) ->
+            session
+                .find(ConfigEntity.class, uuid)
+                .onItem()
+                .ifNotNull()
+                .call(configEntity -> session.merge(configEntity.updateEntity(config)))));
```
ChrisRousey, line 67: Please add your ConfigNotFoundException here

platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+  @Path("{uuid}")
+  public Uni<Response> deleteConfig(@PathParam("uuid") UUID uuid) {
+    return sf.withTransaction(
+        ((session, transaction) ->
+            session
+                .find(ConfigEntity.class, uuid)
+                .onItem()
+                .ifNotNull()
+                .call(configEntity -> session.remove(configEntity))
+                .replaceWith(Response.ok().build())));
```
ChrisRousey, line 80: Please add your ConfigNotFoundException here

platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.List;
+import java.util.UUID;
+import javax.inject.Inject;
+import javax.ws.rs.*;
+import javax.ws.rs.core.MediaType;
+import javax.ws.rs.core.Response;
+import org.eclipse.microprofile.openapi.annotations.parameters.RequestBody;
+import org.hibernate.reactive.mutiny.Mutiny.SessionFactory;
+
+@Path("/config")
```
ChrisRousey, line 15: @flippingbits @HknLof should we always use the plural here?
If so, pipelines actually also only uses the singular form, which we should probably change too

ChrisRousey (changes requested):

olis1996:
> You are using Junit.Jupiter (which is part of junit5) to annotate your tests, but the @before Annotation is for JUnit4. You should use the JUnit5 Annotation `@BeforeAll` and can remove the `import org.junit.Before;` in line 17

It works now. Thanks a lot!

platform-api/src/main/java/io/datacater/core/config/ConfigEndpoint.java
flippingbits: Yes, if we are dealing with resources, like in this case, we should use the plural.

platform-api/src/main/java/io/datacater/core/config/ConfigEntity.java
```diff
@@ -0,0 +1,93 @@
...
+  @JsonProperty("name")
+  private String name;
+
+  @JsonProperty("kind")
+  private String kind;
+
+  @Type(type = JsonTypes.JSON)
+  @JsonProperty("config")
+  @Column(name = "config", columnDefinition = JsonTypes.JSON_BIN)
+  private JsonNode config;
```
flippingbits, line 47: We might want to align these with the other resources, and call it `spec`.

olis1996: I pushed changes that implement your request

olis1996:
I added comments to all test cases, which will be removed before finishing the config endpoint feature, so that my idea behind the test is easier to understand.

flippingbits (dismissed):
Looks good to me 🥳 Feel free to "finalize" the PR once you feel ready

platform-api/src/main/java/io/datacater/core/config/ConfigEntity.java
```diff
@@ -0,0 +1,92 @@
...
+  @UpdateTimestamp
+  @JsonProperty("updatedAt")
+  private Date updatedAt;
+
+  @JsonProperty("name")
+  private String name;
+
+  @JsonProperty("kind")
+  private String kind;
+
```
flippingbits, line 42: It could make sense to add `metadata` as well (similar to [pipelines](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/pipeline/PipelineEntity.java#L48) and others). This would allow us to define labels and, later on,  use other means for referencing configs than UUIDs. 

What do you think, @HknLof ?

HknLof: Yepp, that is good point. @olis1996 Could you get on this?

olis1996: Sure, I'll take care of that!

flippingbits (approved):
Great work, thank you very much!

ChrisRousey (approved):
🔥

HknLof (approved):

PR #79, opened 2022-12-05
Feature/72 format records

### LINKED ISSUE #36: Stream message payload should return a Java-object
as of now, our streammessage object can only accept JSON key/value pairs.
This will work for out customer json and avro serde, because these always resolve to jsonobjects.
Useing other serde (bytearray, etc.) will cause an error.

We should update and test the streammessage with using a java-object instead.

### LINKED ISSUE #72: Support primitive data types for keys and values
At the moment, we only support JSON and AVRO as formats for keys and values.

We should also allow to work with records that use a primitive data type for the key or value, e.g., a String.

flippingbits:
This PR introduces support for Kafka records that hold data in a format that cannot be deserialized to a JSON object.

Fixes #72
Fixes #36

### DIFF
--- a/pipeline/src/main/java/io/datacater/KafkaConfig.java
+++ b/pipeline/src/main/java/io/datacater/KafkaConfig.java
@@ -27,18 +27,18 @@ public class KafkaConfig {
     @Produces
     @ApplicationScoped
     @Identifier("stream-in-configuration")
-    Map<String, Object> streamInConfig() {
+    public static Map<String, Object> streamInConfig() {
         return mapConfig(DATACATER_STREAMIN_CONFIG);
     }
 
     @Produces
     @ApplicationScoped
     @Identifier("stream-out-configuration")
-    Map<String, Object> streamOutConfig() {
+    public static Map<String, Object> streamOutConfig() {
         return mapConfig(DATACATER_STREAMOUT_CONFIG);
     }
 
-    private Map<String, Object> mapConfig(String json){
+    private static Map<String, Object> mapConfig(String json){
         ObjectMapper mapper = new ObjectMapper();
         Map<String, Object> map;
         try{
--- a/pipeline/src/main/java/io/datacater/KafkaConfig.java
+++ b/pipeline/src/main/java/io/datacater/KafkaConfig.java
@@ -19,6 +19,16 @@ public class KafkaConfig {
                     .getOptionalValue("datacater.stream-in.config", String.class)
                     .orElse( "{\"key.deserializer\": \"io.datacater.core.serde.JsonDeserializer\", \"value.deserializer\": \"io.datacater.core.serde.JsonDeserializer\"}");
 
+    static final String DATACATER_STREAMIN_TOPIC =
+            ConfigProvider.getConfig()
+                    .getOptionalValue("mp.messaging.incoming.stream-in.topic", String.class)
+                    .orElse("stream-in");
+
+    static final String DATACATER_STREAMOUT_TOPIC =
+            ConfigProvider.getConfig()
+                    .getOptionalValue("mp.messaging.incoming.stream-out.topic", String.class)
+                    .orElse("stream-out");
+
     static final String DATACATER_STREAMOUT_CONFIG =
             ConfigProvider.getConfig()
                     .getOptionalValue("datacater.stream-out.config", String.class)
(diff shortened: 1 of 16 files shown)


pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -110,17 +141,42 @@ private void logMessage(String message){
     LOGGER.error(errorMsg);
   }
 
-  private JsonObject getKey(JsonObject message){
-    if(message.getJsonObject(PipelineConfig.KEY) == null){
-      return null;
+  private Deserializer getDeserializerInstance(String className) throws TransformationException {
```
HknLof, line 144: It might make sense to map out all of our Deserializers via an Enum.

pipeline/src/main/java/io/datacater/Pipeline.java
```diff
@@ -110,17 +141,42 @@ private void logMessage(String message){
...
+    try {
+      Class deserializerClass = Class.forName(className);
+
+      return (Deserializer) deserializerClass.getConstructor().newInstance();
+    } catch (Exception e) {
+      throw new TransformationException(e.getMessage());
+    }
+  }
+
+  private Serializer getSerializerInstance(String className) throws TransformationException {
```
HknLof, line 154: It might make sense to map out all of our Serializers via an Enum.

serde/src/main/java/io/datacater/core/serde/JsonDeserializer.java
```diff
@@ -4,14 +4,16 @@
 import io.vertx.core.buffer.Buffer;
 import io.vertx.core.json.JsonObject;
 import org.apache.kafka.common.serialization.Deserializer;
+import java.util.Map;
 
-public class JsonDeserializer implements Deserializer<JsonObject> {
+public class JsonDeserializer implements Deserializer<Map<String, Object>> {
```
HknLof, line 9: Shall we rename this to `MapDeserializer`? There might be a point to actually use `JsonObject` internally.

pipeline/src/main/java/io/datacater/KafkaConfig.java
```diff
@@ -27,18 +37,18 @@ public class KafkaConfig {
     @Produces
     @ApplicationScoped
     @Identifier("stream-in-configuration")
-    Map<String, Object> streamInConfig() {
+    public static Map<String, Object> streamInConfig() {
```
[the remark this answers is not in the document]
flippingbits: I believe it should work, all tests are green.

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: At the moment, we allow users to configure any deserializer class.

From my understanding, using an enum would require us to know potential deserializers beforehand.

Are we willing to step into this direction and decrease the flexibility? Happy to make the chance, if I get a yes :)

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: At the moment, we allow users to configure any serializer class.

From my understanding, using an enum would require us to know potential serializers beforehand.

Are we willing to step into this direction and decrease the flexibility? Happy to make the chance, if I get a yes :)

serde/src/main/java/io/datacater/core/serde/JsonDeserializer.java
flippingbits: If we would deserialize to a JsonObject, we would end up with the following string representation:

```
{
  "is_empty": true|false,
  "map": {}
```

We are calling [getMap()](https://vertx.io/docs/apidocs/io/vertx/core/json/JsonObject.html#getMap--) on the `JsonObject` to get the map directly, without the information on emptiness.

I believe this is almost never useful for us. Do you have an use case in mind where we would prefer to work with `JsonObject`?

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: In 2af7b3cc98887f1e60afade8724268585d74ce1f, I replaced the usage of the reflection API with a static Map.

pipeline/src/main/java/io/datacater/Pipeline.java
flippingbits: In 2af7b3cc98887f1e60afade8724268585d74ce1f, I replaced the usage of the reflection API with a static Map.

ChrisRousey (approved):
🔥

Issue #23, opened 2022-11-05
Introduce testing strategy for pre-defined filters and transforms


flippingbits:
We provide a set of pre-defined [filters](https://github.com/DataCater/datacater/tree/main/filters) and [transforms](https://github.com/DataCater/datacater/tree/main/transforms), which is extendable by users.

We should introduce a gold standard for testing the filters and transforms, using a established test framework from the Python community like [pytest](https://docs.pytest.org/en/7.2.x/).

PR #80, opened 2022-12-06
feat(#23) working example of how rls can be implemented

HknLof:
This is a follow up to pull request https://github.com/DataCater/datacater/pull/77. The process to implement RLS for datacater is as follows:
1. Initialise tables and user role `datacater` via migartions. For now `init-db.sql`.
2. Define a `@MappedSuperClass`, which fills in the tenant automatically. [Static for validating approach]
3. On Read queries set the current tenant as a pg param via `SET datacater.tenant = $tenant`. [Static for validating approach]

Refs
- https://github.com/DataCater/datacater/pull/77


flippingbits (approved):
Thanks, LGTM 💯

flippingbits (approved):

flippingbits (approved):

PR #84, opened 2022-12-08
Build: Upgrade quarkus

ChrisRousey:


platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -194,12 +194,13 @@ public void delete(UUID deploymentId) {
...
     Boolean status =
-        client
+        !client
             .apps()
             .deployments()
             .inNamespace(StaticConfig.EnvironmentVariables.NAMESPACE)
             .withName(name)
-            .delete();
+            .delete()
+            .isEmpty();
```
flippingbits, line 203: What is this change about? Is it related to the update?

ChrisRousey: yes, the return type was changed with the upgrade to a list of the deleted items. The check now looks if the list is empty, then nothing was deleted. If it isn't empty, the deployment was deleted successfully

HknLof:
We can close an re-open with the most recent version of main. The startup event is now out of the codebase.

@ChrisRousey  What do you think?

ChrisRousey:
> We can close an re-open with the most recent version of main. The startup event is now out of the codebase.
> 
> @ChrisRousey  What do you think?

Yeah, sounds good. Plus the current version of main has less needed changes since most (or all) needed changes are regarding the k8 client

PR #86, opened 2022-12-09
feat(pipeline): Improve error logging

### LINKED ISSUE #61: Improve logging and error handling of pipeline deployments
Placeholder issue for working on improving the logging behavior of pipeline deployments.

- [ ] If the Python runner sidecar is not available, the pipeline should log an error and throw an exception. It should terminate with an exit code > 0.
- [ ] If one of the records could not be successfully processed by the Python runner, the pipeline should log an error and throw an exception. It should NOT send the record to the stream out. If a dead letter queue is defined, the record should be sent to the dead letter queue. If no dead letter queue is defined, the processing should stop, allowing the developer to skip the record by resetting the offset of the pipeline or to continue the processing by adjusting the pipeline in a way such that it can process the record without throwing errors. At the moment, we do not support a dead letter queue, thus I suggest to start with the latter part.

flippingbits:
When the processing of a record fails, log the following information:
* key of the record
* value of the record
* metadata of the record (e.g., offset)
* detailed information on the error (e.g., exception name, exception message, stacktrace, etc.)

Fixes #61

### DIFF
--- a/pipeline/src/main/java/io/datacater/Pipeline.java
+++ b/pipeline/src/main/java/io/datacater/Pipeline.java
@@ -7,7 +7,6 @@ import io.vertx.core.json.JsonObject;
 import io.smallrye.reactive.messaging.kafka.Record;
 
 import java.time.Duration;
-import java.util.UUID;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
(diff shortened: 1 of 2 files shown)


ChrisRousey (approved):
LGTM. Thank you for fixing this

Issue #76, opened 2022-11-28
Support installing DataCater into non-default namespaces


flippingbits:
At the moment, our `UI` pod is hardly wired with the `default` namespace. When trying to install DataCater into a non-`default` namespace, it will crash with the following error message:

```
nginx: [emerg] host not found in upstream "datacater.default.svc.cluster.local" in /etc/nginx/conf.d/default.conf:20

flippingbits:
Fixes with e35b932f4b40e3ccb667dfe1566fbdfce883a609

PR #88, opened 2022-12-13
feat(nginx): user service name for proxy pass

HknLof:
This change introduces an ui service, deployment, and
changes to the nginx conf. With this change we are
going towards the ability of installation in any namespaces.

Refs:

- https://github.com/DataCater/datacater/issues/76


flippingbits (approved):
LGTM, thanks a lot for providing this important enhancement! :)

PR #89, opened 2022-12-14
Do not let DataCater create Kubernetes namespaces

flippingbits:
The creation of Kubernetes namespaces should not be performed by DataCater but handled by tools like Helm.

This PR removes the functionality where DataCater automatically creates namespaces for the Python Runner and Deployments. This allows us to remove the cluster role.


HknLof (approved):
Thank you for the change. I will merge and deploy locally for validating your changes.

PR #93, opened 2022-12-16
feat(ui): Sort resources by their update time

flippingbits:
In the UI, sort resources by their `updatedAt`field, showing the most recently updated resources first in the list.

This helps users to navigate their resources in a more transparent way, since names and UUIDs are sometimes not enough for differentiating different resources.

We now also show the update time in the UI:

Screenshot 2022-12-16 at 10 05 36

### DIFF
--- a/ui/package-lock.json
+++ b/ui/package-lock.json
@@ -10,6 +10,7 @@
       "dependencies": {
         "axios": "^0.27.2",
         "bootstrap": "^5.2.1",
+        "javascript-time-ago": "^2.5.9",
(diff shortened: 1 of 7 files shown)


ChrisRousey (changes requested):
Great idea with the sorting. Looks good, but could the "last changed date"-position differs from resource to resource. I think it would make sense, to unify this.

flippingbits:
Thanks, you are right. Timestamps are now positioned the same way for all resources.

ChrisRousey (approved):
Thank you, looks great

PR #97, opened 2022-12-19
feat(ci): Build latest images for main branch

flippingbits:
For the most recent commit in the `main` branch, provide images with the tag `latest`.


HknLof:
Could you add `quarkus.container-image.additional-tags` with the commit hash?

[Configuration optsions https://quarkus.io/guides/all-config#quarkus-container-image_quarkus.container-image.additional-tags]

That way we can trace what commit is running or what the image is actually targeting. Further, I would prefer main to be `nightly` and releases to be latest.

My reasoning is, that Kubernetes and Docker fill in `latest`, if no tag was given. [Last time I checked]
What do you think?

flippingbits:
Makes sense, thanks for your feedback :) 

I renamed `latest` to `nightly` and additionally tagged all nightly images (not only the Quarkus ones) with their SHA.

HknLof (approved):

Issue #20, opened 2022-11-03
Come up with reasonable default resource requests for pipelines


ChrisRousey:
As of now, we configure each pipeline deployment to use 1.5GB of memory.

With this issue, we should evaluate how much memory should be allocated to each deployment as a standard.

flippingbits:
From my experience, we can reduce the requested memory to 300MB for non-native pipelines, and might be able to further reduce the requests for native pipelines.

WDYT, @ChrisRousey @HknLof

Issue #15, opened 2022-10-26
Make Kubernetes Deployment Configuration configurable


ChrisRousey:
As of now, we are using a static K8 configuration for things such are resources (replicas, memory, cpu...). 
The Static configurations can be found in the `Deployment.StaticConfig`-class.
The Configurations are being set while creating a deployment in the `Deployment.K8Deployment.create`-method

The goal of this ticket is to use a dynamic config when starting a deployment.
For this, we should: 
- have standard values in the case that user doesn't explicitly set them (we can use the current static values as a default).
- create a new class `DeploymentSpec` should be introduced to hold the configuration
  * An object of this class can optionally be given to put/post deployment
- We should use the k8 naming for the configurations i.e. `replicas`, `resources.requests.memory`, `resources.limits.memory`...

HknLof:
Setting Replicas is still open.

PR #98, opened 2022-12-19
feat(deployment): make resources configurable

ChrisRousey:
this PR makes resource limits and requests for deployments configurable and lowers our default value.

refs: #20, #15

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/deployment/StaticConfig.java
+++ b/platform-api/src/main/java/io/datacater/core/deployment/StaticConfig.java
@@ -24,11 +24,11 @@ public class StaticConfig {
   static final Map<String, Quantity> RESOURCE_REQUESTS =
       Map.of(
           "cpu",
-          new Quantity(EnvironmentVariables.DEPLOYMENT_RESOURCES_CPU_REQUEST),
+          new Quantity(EnvironmentVariables.DEPLOYMENT_RESOURCES_REQUEST_CPU),
           "memory",
-          new Quantity(EnvironmentVariables.DEPLOYMENT_RESOURCES_MEMORY_REQUEST));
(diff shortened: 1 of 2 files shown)


flippingbits (commented):
Looks good to me, thanks a lot :)

Only thing I would change is making the configuration 100% compatible with [Kubernetes](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container), .i.e., instead of `datacater.deployment.resources.cpu.request` we would have `datacater.deployment.resources.requests.cpu`.

This way we help users that are new to DataCater but have experience with Kubernetes.

ChrisRousey:
> Only thing I would change is making the configuration 100% compatible with [Kubernetes](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#resource-requests-and-limits-of-pod-and-container), .i.e., instead of `datacater.deployment.resources.cpu.request` we would have `datacater.deployment.resources.requests.cpu`.

Good point, thanks! I just pushed the change :)

platform-api/src/main/java/io/datacater/core/deployment/StaticConfig.java
```diff
@@ -112,6 +117,19 @@ private EnvironmentVariables() {}
         ConfigProvider.getConfig()
             .getOptionalValue("datacater.pythonrunner.image.containerPort", Integer.class)
             .orElse(50000);
+
+    static final String DEPLOYMENT_RESOURCES_REQUEST_MEMORY =
+        ConfigProvider.getConfig()
+            .getOptionalValue("datacater.deployment.resources.request.memory", String.class)
```
flippingbits, line 123: Can you please make `request` plural, .i.e., `requests`, to match the `application.yml`?

flippingbits (commented):
Thanks :) Found one small issue

flippingbits (approved):
Thanks, LGTM 💯

PR #99, opened 2022-12-20
Pass stream configs as environment variables to pipelines

flippingbits:
We cannot use Smallrye's kafka-configuration option for passing channel configs to pipelines that are running in the native mode.

To support native mode, this PR introduces the usage of environment variables for configuring streams/channels of the pipeline.

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
+++ b/platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
@@ -323,6 +323,23 @@ public class K8Deployment {
         StaticConfig.VALUE_SERIALIZER,
         getEnvVariableFromNode(streamOut.getSpec(), StaticConfig.VALUE_SERIALIZER));
 
+    // Store Serde-related information
+    Map<String, Object> dataCaterConfig = new HashMap<>();
+    dataCaterConfig.put(
+        StaticConfig.KEY_DESERIALIZER, streamInConfig.get(StaticConfig.KEY_DESERIALIZER));
+    dataCaterConfig.put(
+        StaticConfig.VALUE_DESERIALIZER, streamInConfig.get(StaticConfig.VALUE_DESERIALIZER));
+    dataCaterConfig.put(
+        StaticConfig.KEY_SERIALIZER, streamOutConfig.get(StaticConfig.KEY_SERIALIZER));
+    dataCaterConfig.put(
+        StaticConfig.VALUE_SERIALIZER, streamOutConfig.get(StaticConfig.VALUE_SERIALIZER));
+
+    // Remove Serde-related information from channel configs
+    streamInConfig.remove(StaticConfig.KEY_DESERIALIZER);
+    streamInConfig.remove(StaticConfig.VALUE_DESERIALIZER);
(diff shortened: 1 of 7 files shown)


platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -318,24 +323,85 @@ private List<EnvVar> getEnvironmentVariables(
...
+    // Store Serde-related information
+    Map<String, Object> dataCaterConfig = new HashMap<>();
+    dataCaterConfig.put(
+        StaticConfig.KEY_DESERIALIZER, streamInConfig.get(StaticConfig.KEY_DESERIALIZER));
+    dataCaterConfig.put(
+        StaticConfig.VALUE_DESERIALIZER, streamInConfig.get(StaticConfig.VALUE_DESERIALIZER));
+    dataCaterConfig.put(
+        StaticConfig.KEY_SERIALIZER, streamOutConfig.get(StaticConfig.KEY_SERIALIZER));
+    dataCaterConfig.put(
+        StaticConfig.VALUE_SERIALIZER, streamOutConfig.get(StaticConfig.VALUE_SERIALIZER));
```
HknLof, line 335: Is there a reason we do not use `Map.of(K1, V1, K2, V2, ...)` for initiating the `Map` `datacaterConfig`?

HknLof (commented):
From an implementation logic perspective your PR looks good. If there is a reason why we refrain from using `Map.of`, then I am happy with the all around implementation.

https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/Map.html#of()

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -318,24 +323,85 @@ private List<EnvVar> getEnvironmentVariables(
...
+   * <p>Examples: - `datacater.name` becomes `DATACATER_NAME` - `datacater.middle-level.name`
+   * becomes `DATACATER_MIDDLE_LEVEL_NAME`
+   *
+   * @param configName
+   * @return config name as valid environment variable
+   */
+  private String transformConfigOptionToEnvVariable(String configName) {
+    return configName
+        // Replace all characters, except a-z A-Z and underscores, with an underscore
+        .replaceAll("[^a-zA-Z_]", "_")
```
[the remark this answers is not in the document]
flippingbits: Yes, this would help a lot. I am also not a fan of it.

AFAIK, the Kafka consumer and producer APIs do not use any properties with a `dash` in the name, so this should not cause any harm.

Nonetheless I still prefer to keep this part for making sure that we do not pass any invalid notation to the environment variables. I am aware of libraries that use it, e.g., [Smallrye](https://quarkus.io/guides/kafka#configuring-smallrye-kafka-connector).

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
flippingbits: Nope. Thanks for the hint! Fixed with 985e148bdc9efbf4745d603baa2faf1d9b49254a

HknLof (approved):

HknLof (approved):

PR #101, opened 2022-12-20
fix(ui): Highlight filters

flippingbits:
If a pipeline step only defines a filter but no transform, highlight its name in the grid view.
This helps users to quickly navigate pipelines.

Before:

Screenshot 2022-12-20 at 20 39 30

After:

Screenshot 2022-12-20 at 20 39 52


HknLof:
Can I stilly apply transformations? I do not see this option above the field `name` any longer!

flippingbits:
Yes, you can. Just click on the "Transform" tab in the context bar. If a transform is applied, we will highlight the transform instead of the filter in the header of the table.

ChrisRousey (approved):
makes sense and looks/works good 👍🏼

flippingbits:
Thanks!

PR #102, opened 2022-12-20
fix(ui): Close sidebar correctly when resetting transform

flippingbits:
When resetting a transform or filter, close the sidebar correctly and do not leave an empty area.

The _empty area_ is clearly visible on the following screenshot:

Screenshot 2022-12-20 at 20 52 30


HknLof (approved):

HknLof:
Would it be possible to centralise the four changes?

flippingbits:
I don't think so. Sorry

PR #103, opened 2022-12-21
feat(authorization): runtime authorization switcher

ChrisRousey:


ChrisRousey:
@HknLof i added everything we need for the PR and everything works as expected, except that a random `io.quarkus.vertx.http.runtime.security.AuthenticationRedirectException` is thrown on authentication, even though the authentication and redirection are working correctly

ChrisRousey:
working on the tests right now fyi.
Tests are failing due to authentication with endpoints. I'm not 100% sure where the catch is though. Tests aren' being authorized.

I can turn the auth off, then the only tests failing are the authorization endpoint tests. My guess is a jwt token is being created and cached during the tests which isn't working correctly

HknLof (approved):

PR #104, opened 2022-12-21
feat(ui/pipeline-designer): Allow changing sample size

flippingbits:
Allow users to change the size of the sample set in the pipeline designer, which is set to 100 by default.

Use a similar approach as used for the inspection of streams:

### DIFF
--- a/ui/src/containers/pipelines/EditPipeline.js
+++ b/ui/src/containers/pipelines/EditPipeline.js
@@ -47,6 +47,7 @@ class EditPipeline extends Component {
       currentStep: undefined,
       debugRecord: undefined,
       errorMessage: "",
+      inspectLimit: 100,
       pipeline: {},
       pipelineUpdated: false,
       pipelineUpdatedAt: undefined,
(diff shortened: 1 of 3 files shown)


ui/src/containers/pipelines/EditPipeline.js
```diff
@@ -676,6 +678,20 @@ class EditPipeline extends Component {
     });
   }
 
+  updateInspectLimit(event) {
```
ChrisRousey, line 681: (idk if this is the correct code for what i'm commenting)
This updates every character event. So if the field is empty and i want 100 records it will first filter to 1, then to 10 then to 100 and each time it filters it can take a few seconds to load.

I think it would be more user friendly if we updated on enter instead of events?

flippingbits: Good catch, I agree! 👍

flippingbits: Fixed with 2fc78b3

ChrisRousey (approved):
it's working smoothly now, thanks 🙂

PR #109, opened 2022-12-22
fix(ui): Make log parsing more robust

flippingbits:
By default, the `/deployments/$uuid/logs` endpoint returns a deployment's logs in the JSON-LD format.

In rare cases, some lines are simple strings not JSON objects. This commit allows us to deal with such edge cases without throwing an error in the frontend.

### DIFF
--- a/ui/src/containers/deployments/DeploymentLogs.js
+++ b/ui/src/containers/deployments/DeploymentLogs.js
@@ -162,13 +162,28 @@ class DeploymentLogs extends Component {
                       >
                         {this.state.logMessages.map((logMessage, idx) => (
                           <React.Fragment key={idx}>
-                            <span className="me-2" style={{ color: "#ae75e6" }}>
-                              {logMessage.timestamp}
-                            </span>
-                            <span className="me-2" style={{ color: "#b4d2ea" }}>
-                              {logMessage.loggerName}
-                            </span>
-                            {logMessage.message} <br />
+                            {typeof logMessage === "string" && (
+                              <>
+                                {logMessage} <br />
+                              </>
+                            )}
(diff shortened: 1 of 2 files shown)


ui/src/containers/deployments/DeploymentLogs.js
```diff
@@ -162,13 +162,28 @@ class DeploymentLogs extends Component {
...
+                                </span>
+                                <span
+                                  className="me-2"
+                                  style={{ color: "#b4d2ea" }}
+                                >
+                                  {logMessage.loggerName}
+                                </span>
+                                {logMessage.message} <br />
+                              </>
+                            )}
```
HknLof, line 186: I think, there is a lot of implicit things happening.

My understanding:

1. Try to parse the logline as json in `deployments.js`.
2. Because of this we assume `logMessage` is of type `string` or `object`.
3. If not `string` it is `object`, we can access `message`, `loggerName`, etc. as properties. This might break, depending on return type of `deployment.js`

We should clarify in `deployment.js`, that the only acceptable return types are `string` and `object` with the properties `timestamp, loggerName, and message`.

HknLof (changes requested):

ui/src/containers/deployments/DeploymentLogs.js
flippingbits: I agree, this should only be a temporary fix.
Once we addressed issue #45, we can finally remove the JSON parsing again.

flippingbits: As discussed, let's rather focus on fixing issue #45. Closing the PR now. Thanks! :)

PR #113, opened 2023-01-02
feat(api-path): remove root-path from non-app routes

ChrisRousey:
This PR is to remove our versioned path prefix from non-application resources that are not managed by us (`/health`, `/metrics`, etc.)

### DIFF
--- a/platform-api/src/main/resources/application.yml
+++ b/platform-api/src/main/resources/application.yml
@@ -26,6 +26,7 @@ quarkus  :
           admin: admin,dev,user
   http:
     root-path: /api/${datacater.api.version}
+    non-application-root-path: /q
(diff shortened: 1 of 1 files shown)


ChrisRousey:
@flippingbits @HknLof 
By settings the non-application path, our root-path is ignored for these resources (this also includes things like `/q/dev`). I set the base non-application-root-path to `/q`, as thisis the quarkus standard. We could change this to something we find more fitting. Any suggestions?

flippingbits (approved):
This was more straightforward than I thought, thanks :)

**We must not forget that, with the next release, we must update the endpoints for the health and readiness probes in our Helm charts and Kubernetes manifests.**

PR #114, opened 2023-01-03
feat(ui/streams): Update limit on blur

### LINKED ISSUE #106: Update sample records limit only on blur or when pressing Enter
Similar to the pipeline designer, when inspecting a stream and changing the number of the sample records, we should persist a change only on blur or when pressing _Enter_, not on every keystroke.

cc @ChrisRousey

flippingbits:
When inspecting a stream in the UI, update the limit parameter on blur and not on change, to provide a more convenient way for entering multi-digit numbers.

Fixes #106


ChrisRousey (approved):
Works great, thank you 🙂

Issue #67, opened 2022-11-22
Stateful set python-runner does not get cleaned up


flippingbits:
When uninstalling DataCater via `kubectl delete -f ...`, all DataCater resources except the `python-runner` stateful set get removed.

We should fix this to leave a clean Kubernetes namespace after removing DataCater.

HknLof:
Is there a Quarkus / Kubernetes event, that we might use for this? Otherwise, let us remove the initiation of our `python-runner` StatefulSet from the application and build it into helm.

flippingbits:
This should be fixed now, closing the issue.

PR #115, opened 2023-01-04
feat(python-runner): remove python-runner pool creation from the app

ChrisRousey:
This PR removes the python-runner statefulset pool creation from inside the application on startup. 

We will administrate kubernetes resources inside kubernetes, instead of the application. This makes sure the python-runner pool is terminated when the application is terminated

refs: #67


HknLof:
@ChrisRousey Could you resolve the merge conflict locally?

HknLof (approved):

PR #116, opened 2023-01-04
feat(helm-chart): administrate python-runner with helm

ChrisRousey:
This PR changes the helm chart, so the python-runner pool is administrated with k8 instead of inside the application

refs: #67


helm-charts/datacater/templates/pythonrunner.yaml
```diff
@@ -0,0 +1,45 @@
+apiVersion: apps/v1
+kind: StatefulSet
+metadata:
+  name: {{ .Values.datacater.pythonrunner.name }}
+  labels:
+    {{- include "datacater.labelsPythonRunner" . | nindent 4 }}
+spec:
+  serviceName: {{ .Values.datacater.pythonrunner.serviceName }}
```
HknLof, line 8: This can not be variable, if it is not set to `pythonrunner-service.yaml#L4`. The service itself and this name have to be equal.

HknLof: I think fixing these names is better than keeping them variable. Our application has also assumptions and we can only change these values at startup time.

HknLof (commented):
We need to be sure to really have no naming mismatches. We can either make the expected name and label of the python runner configurable and reuse everywhere. Including the following line or we fix these values to a certain extent.

https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/kubernetes/DataCaterK8sConfig.java#L16

HknLof (dismissed):

HknLof (approved):

PR #118, opened 2023-01-09
When previewing a pipeline in the UI, visualize changed records

### LINKED ISSUE #42: [UI] Highlight fields/cells that are changed by the current step
In the pipeline designer, we should visually highlight the cells (fields of rows) that are changed by the current step, as done in our "old" platform:

Screenshot 2022-11-16 at 13 02 29

flippingbits:
Let the `/preview` endpoint return which records have been changed by a pipeline:

* Each record's `metadata` now feature a new field `lastChange`
* The metadata field `lastChange` holds two keys, `key` and `value`
* If the record key holds JSON data, the object `record.metadata.lastChange.key` features an entry for each changed field of the JSON object, indicating the last step that has changed the field
* If the record value holds JSON data, the object `record.metadata.lastChange.value` features an entry for each changed field of the JSON object, indicating the last step that has changed the field

Example record:

```json
{
        "key": null,
        "value": {
                "name": "DataCater GmbH",
                "website": "https://datacater.io"
        }
}
```

If the previewed pipeline modifies the value's field `name` in the third step, the response of the `/preview` endpoint looks as follows:

```json
{
        ...,
        "metadata": {
                ....,
                "lastChange": {
                        "key": {},
                        "value": {
                                "name": 2
                        }
                }
        }
}
```

The UI can use these information to visually highlight cells/fields that have been modified by the previewed pipeline:

Screenshot 2023-01-09 at 13 27 41

Fixes #42

### DIFF
--- a/python-runner/runner.py
+++ b/python-runner/runner.py
@@ -1,16 +1,17 @@
 # Start app: $ python3 -m uvicorn runner:app --log-level critical
+import copy
 import datetime
-
+import json
 import re
 import sys
 import traceback
+import yaml
+
 from glob import iglob
 from importlib import import_module
 from os import path
-from typing import List, Optional, Any
 
-import json
-import yaml
+from typing import List, Optional, Any
 from fastapi import FastAPI, Response, status, Request
 from fastapi.responses import JSONResponse
 from pydantic import BaseModel
@@ -78,7 +79,7 @@ if path.isfile(CONFIG_MAP_FILE):
     config_map.close()
 
 
-def apply_pipeline(record: dict, pipeline: dict, preview_step=None):
+def apply_pipeline(record: dict, pipeline: dict):
     location = {}
     try:
         step_index = 0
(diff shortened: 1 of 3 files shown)


python-runner/runner.py
```diff
@@ -78,11 +79,133 @@ class PipelineRequest(BaseModel):
     config_map.close()
 
 
-def apply_pipeline(record: dict, pipeline: dict, preview_step=None):
+def apply_pipeline(record: dict, pipeline: dict):
```
HknLof, line 82: Are these changes intended to be part of this PR?

python-runner/runner.py
```diff
@@ -192,40 +315,57 @@ def apply_pipeline(record: dict, pipeline: dict, preview_step=None):
...
-                "lineNumber": line_number,
-            },
-            "stacktrace": formatted_errors,
-        }
-
-        return record
+        return add_error_to_metadata(record, location, sys.exc_info()[0], traceback)
+
+
+def add_error_to_metadata(record, location, error_type, traceback):
```
HknLof, line 341: Would be ok to have this in this PR. But we should consider visualising errors on a field level. Not sure if this is possible this way.

HknLof (commented):
Left a couple of comments, that we should clarify :) For the majority it looks good.

python-runner/runner.py
flippingbits: Yes. This PR introduced the new method `preview_pipeline`, which is only used for the `/preview` endpoint. It performs a couple of tasks that are not needed for the deployment and might have a slightly negative impact on the performance.

flippingbits:
Thanks, I hope I could clarify things.

python-runner/runner.py
HknLof: Correcting my previous comment. We should be able to visualise errors on a cell level.

flippingbits: This is still possible since this metadata belongs to a specific record.

HknLof (approved):

PR #120, opened 2023-01-09
fix(helm): introduce some defaults for db connections and fix python-runner reference

HknLof:
@flippingbits @ChrisRousey Found some errors on a fresh install. Once, tested with new feature, I will update the `minikube-any-namespace.yaml`.

### DIFF
--- a/helm-charts/datacater/values.yaml
+++ b/helm-charts/datacater/values.yaml
@@ -22,7 +22,7 @@ datacater:
       repository: "datacater/python-runner"
       pullPolicy: IfNotPresent
       version: "nightly"
-    replicas: 10
+    replicas: 5
     serviceName: "pythonrunner"
     resources:
       requests:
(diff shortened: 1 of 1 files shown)


helm-charts/datacater/values.yaml
```diff
@@ -19,9 +19,9 @@ datacater:
   pythonrunner:
     name: "python-runner"
     image:
-      name: "datacater/python-runner"
+      repository: "datacater/python-runner"
       pullPolicy: IfNotPresent
-      version: "nightly-20230105"
+      version: "nightly"
```
flippingbits, line 24: We use `nightly` as our `latest`-tag. Is the update of the version to `nightly` intentional despite using a specific nightly build for [other components](https://github.com/DataCater/datacater/blob/main/helm-charts/datacater/values.yaml#L14)?

flippingbits (commented):
I have one open question :)

helm-charts/datacater/values.yaml
HknLof: Will fix it. The reason was, that minikube could not find that tag. Slipped through :)
It is available, though https://hub.docker.com/r/datacater/python-runner/tags?page=1&name=nightly-20230105.

flippingbits (approved):
Thank you! :)

PR #122, opened 2023-01-10
Feat(k8-manifest): introduce a manifest containing a redpanda broker

ChrisRousey:
Added a manifest which also deploys a redpanda broker.

To access the broker inside datacater, `bootsrap.servers` is `redpanda-0.redpanda.default.svc.cluster.local.:9093`
To produce data to broker, use the following command: 
`kubectl exec -it -n default redpanda-0 -- rpk topic produce topicName --brokers='redpanda-0.redpanda.default.svc.cluster.local.:9093'`

Currently running minikube with:
- cpus: 4
- memory: 8192

## Update
By editing `/etc/hosts/` and adding the line `127.0.0.1 redpanda-0.redpanda.default.svc.cluster.local.`, you can then portforward `kubectl port-forward redpanda-0 9093:9093` and access the topics through `localhost`. This is much faster to produce test records, than using rpk inside the cluster

### DIFF
--- a/k8s-manifests/README_RedPanda.md
+++ b/k8s-manifests/README_RedPanda.md
@@ -1,0 +1,52 @@
+# Create and working with Redpanda manifest
+Redpanda offers a Helm chart for a quick and easy installation.
+We convert this helm chart to a Kubernetes Manifest to provide a quick and easy 
+installation and integration with our Open-Core product.
+If the redpanda version needs to be updated, the following steps can be done.
+
+1. Add the RedPanda repository to your helm client
+```
+helm repo add redpanda https://charts.redpanda.com/
+```
+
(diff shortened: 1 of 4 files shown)


flippingbits:
Thanks, that is a very helpful for testing and playing with DataCater! :)

Can we somewhere document how to generate these manifests, so we can easily update them if needed?

HknLof (dismissed):

ChrisRousey:
@HknLof @flippingbits i'll update the manifests labels in this PR and add a readme how to work with and use redpanda in cluster

Should i also add the changes to run pythonrunner stateful set in k8 instead of through the application? The changes for this are already in main but no in the currently referenced image `nightly-20230105`

HknLof (approved):

PR #124, opened 2023-01-11
fix: minor fixes to Streams and SerDe

ChrisRousey:
This PR fixes 2 bugs:
- SerDe errors when key or value are null
- Stream not using the partitions from as defined


ChrisRousey:
I will also look into another bug in the PR.
The replication factor can't be higher than the amount of available brokers. If we set this higher, the user doesn't get any feedback and the topic is silently not created. The topic is created in DC though

ChrisRousey:
> LGTM, Thank you! :)
> 
> Can you please add the `null` protection to the Avro SerDes, too?

Avro already has null catch's
[Serializer](https://github.com/DataCater/datacater/blob/main/serde/src/main/java/io/datacater/core/serde/AvroSerializer.java#L55-L57), [Deserializer](https://github.com/DataCater/datacater/blob/main/serde/src/main/java/io/datacater/core/serde/AvroDeserializer.java#L61-L63)

flippingbits (approved):

PR #127, opened 2023-01-12
Switch to js-yaml for better YAML formatting

flippingbits:
This PR replaces the existing JSON-to-YAML formatter with `js-yaml`.

The main intention behind this PR is to render multiline strings in a clean way.

Before:
Screenshot 2023-01-12 at 09 24 53

After:
Screenshot 2023-01-12 at 09 22 28


flippingbits:
My apologies for the forced push. I had to rewrite the Git history since I accidentally committed a file that does not belong here 😰

HknLof (approved):

PR #152, opened 2023-01-31
feat(helm): Upgrade to 603ed4

flippingbits:
Upgrade the Helm chart and Kubernetes manifests to `603ed45f1d05c69ab75179fd6f78455f398967bb` to fix a couple of issues with backpressure in the pipelines.

### DIFF
--- a/helm-charts/datacater/Chart.yaml
+++ b/helm-charts/datacater/Chart.yaml
@@ -25,4 +25,4 @@ maintainers:
   - name: DataCater
     url: https://datacater.io
 
-appVersion: "2023.1"
+appVersion: "603ed45f1d05c69ab75179fd6f78455f398967bb"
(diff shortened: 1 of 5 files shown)


HknLof (commented):
Were these changes via search and replace?

flippingbits:
Yes, did I miss something?

HknLof:
Not that I can see, but I think for consistency purposes, we should aim to generate `k8s-manifests/minikube-any-namespace.yaml` via an helm command. So that we stay consistent with installation paths.

Below is a command. It should work without the two postgres flags.

https://github.com/DataCater/datacater/blob/main/k8s-manifests/README.md#create-plain-datacater-minikube-manifest

flippingbits:
I agree 👍  Will update the manifest in a bit.

Are you aware of the command that we used to create `k8s-manifests/minikube-with-postgres-ns-default.yaml`?

HknLof:
I think the command should be the same with the additional flag `-n default`.

HknLof (changes requested):

HknLof (approved):

PR #154, opened 2023-02-02
fix: reference mount path to pg

HknLof:
The directory for `PGDATA` has to be a sub-directory of the mount path.

Refs:
- https://hub.docker.com/_/postgres


flippingbits (dismissed):
LGTM, thanks for the fix

flippingbits (approved):

PR #156, opened 2023-02-09
fix(ui): Fix previewing tombstone records

flippingbits:
Fix the previewing of tombstone records, i.e., records whose value is `null`, by rendering all fields as `null` in the grid.

### DIFF
--- a/ui/src/components/pipelines/pipeline_designer/Grid.js
+++ b/ui/src/components/pipelines/pipeline_designer/Grid.js
@@ -23,7 +23,8 @@ class Grid extends Component {
 
     const field = column.field;
     const fieldName = field.name;
-    const rawValue = sample["value"][fieldName];
+    const rawValue =
+      sample["value"] == null ? null : sample["value"][fieldName];
 
     // Check whether an error has occured when applying the pipeline spec
     // in the current or a previous step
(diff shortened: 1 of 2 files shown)


ChrisRousey:
Bildschirmfoto 2023-02-09 um 19 48 54

Doesn't seem to be fixed.
Sample dataset i used is shown like this with rpk:
```
{
  "topic": "test-partitions",
  "key": "{\"record_key\": 9999}",
  "value": "null",
  "timestamp": 1675968141752,
  "partition": 4,
  "offset": 11927
}
```

The following sample doesn't produce an error, but it isn't `null`, it's empty:
```
{
  "topic": "test-partitions",
  "key": "{\"record_key\": 9994}",
  "value": "{}",
  "timestamp": 1675968686280,
  "partition": 2,
  "offset": 9872
}
```

flippingbits:
This looks more like an error that is being thrown by the deserializer. It looks like you are trying to interpret the string(!) `"null"` as JSON?

I was trying to fix a UI bug that occurred with tombstone records, e.g.:

```
{
  "topic": "test-partitions",
  "key": "{\"record_key\": 9994}",
  "value": null,
  "timestamp": 1675968686280,
  "partition": 2,
  "offset": 9872
}

ChrisRousey (approved):
Looks good. The error i had came from the python serializer i was using. It converted null into a null String, which couldnt be interperted as a json

flippingbits:
Thanks for investigating and clarifying the issue and of course for the review! :)

PR #157, opened 2023-02-09
Feature: config reference mapping

ChrisRousey:
This PR implements configuration referencing for top-level resources to provide the users of Datacater with reusable and collaborative configurations


ChrisRousey:
Thanks for the Feedback on streams. From the conversation this morning, we still have a few pain-points which need to be addressed in this PR.

- We need a different way for matching objects and configs.
- We need more transparancy for Resources (streams, deployments) when they use config options, and a config is updated. 
- We need to create the possability to refence multiple configs for each resource.

ChrisRousey:
We need a different way for matching objects and configs:

We will go forward by using a k8 style selector for configs and resources. Configs should have a selector label in the metadata node `app.datacater.io/name` and resources should refence these labels for matching.

ChrisRousey:
We need more transparancy for Resources (streams, deployments) when they use config options, and a config is updated:


- As of now, we will add a versioning to configs. This means, when updating a config, we will not overwrite it but add a new config object with a newer version of the old config. 
- In turn, streams must also reference a version of a config. 
    - Users will not be able to change the version, the most recent version will always be used.
    - When updating a stream, the config version will be automatically updated if there is a newer version.
- This allows us to have a reference to a config at the time of a stream creation, not losing any information but also allowing us to retain the original stream object definition without overwriting properties.

ChrisRousey:
We need to create the possibility to refence multiple configs for each resource:

- Streams should be able to reference multiple configs.
    - Configs could separately hold connection and config information for example.
- When multiple config versions are found, the most recent will be used and the stream config reference will be updated
- When multiple config references are found, we will try and match them together. If multiple configs contains the same property, we will fail and return the information to the user.
    - it was discussed to add a priority field to configs. I don't see this as an optimal implementation seeing as that we could still have configs with the same prio and containing the same config properties. Using priorities only adds an extra level of maintenance for the users while only moving the problem from one area to another.

ChrisRousey:
It is still up for discussion whether to Prioritize stream or config properties.

- We currently Prioritize properties coming from configs, so we overwrite stream fields.
    - We consider configs to be at a higher level, since configs are used to change streams

- It could also be argued, that streams should have a higher priority than configs.
    - Streams are "aware" of having configs, meaning we _could_ expect streams to be the source of truth and configs only considered to be an addon.


*streams are used as an example, this information also applies to other resources

ChrisRousey:
@flippingbits @HknLof I have added some information from our discusison this morning. Feel free to add information where you see needed and don't hesitate if you see any problems/errors in the proposed approaches. Thanks 🙂

ChrisRousey:
> To be on the safe side 🤞 , can you please clarify `for now the information would go lost`?

When working with streams/deployments inside this PR (and after merging until the versioning pr is finished) the exact information regarding which config was used for the creation of the given resource will not be saved anywhere.

flippingbits:
> When working with streams/deployments inside this PR (and after merging until the versioning pr is finished) the exact information regarding which config was used for the creation of the given resource will not be saved anywhere.

Thank you for the clarification! 👍  That is not optimal but it is something that we could - from my point of view- walk with until we have the versioning in place. In the meantime, we could add log statements that help with debugging config/stream integrations. What do you think @ChrisRousey @HknLof?

ChrisRousey:
@HknLof @flippingbits 
Regarding the label matching:
- We have a labels field in deployments and streams defined as Map<String, List<String>> to reference more than one item with the same label.
- When creating a config, we automatically create a label in the Metadata mapping the config name to the label 'app.datacater.io/name'
- Config MetaData is saved as a JsonNode in Postgres. To simplify the matching when looking for a referenced config, instead of searching through the metadata, i have mapped it to only look for the referenced name.

Does this sound ok?
I also have a question regarding the label naming.
- Config labels are mapped to a name label -> 'app.datacater.io/name'
- should the resources (stream/deployment) label also be name or rather 'app.datacater.io/config'?

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
```diff
@@ -0,0 +1,118 @@
...
+                .getResultList()
+                .onItem()
+                .transform(
+                    x -> x.stream().filter(item -> configs.contains(item.getName())).toList())
+                .onItem()
+                .ifNull()
+                .continueWith(new ArrayList<>()));
+  }
+
+  public static List<String> getConfigNames(Map<String, List<String>> labels) {
```
flippingbits, line 29: I feel that the function name should indicate that it is searching the labels and not the names of the configs.

It looks like it is always returning one name, so maybe make the name singular?

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
```diff
@@ -0,0 +1,118 @@
...
+    // TODO consider other label options
+    if (labels != null) {
+      return labels.get("app.datacater.io/name");
+    }
+    return new ArrayList<>();
+  }
+
+  public static Stream combineWithStream(Stream stream, List<ConfigEntity> configList) {
```
flippingbits, line 37: Would a name like `applyConfigToStream` or `enhanceStreamWithConfig` better indicate what is being done here?

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
```diff
@@ -0,0 +1,118 @@
...
+    if (!configList.isEmpty()) {
+      validateConfigList(configList);
```
flippingbits, line 39: Is checking for duplicate label keys the only validation performed? Then you should explicitly name it.

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
```diff
@@ -0,0 +1,118 @@
...
+  public static List<String> getConfigNames(Map<String, List<String>> labels) {
+    // TODO consider other label options
+    if (labels != null) {
+      return labels.get("app.datacater.io/name");
+    }
+    return new ArrayList<>();
+  }
+
+  public static Stream combineWithStream(Stream stream, List<ConfigEntity> configList) {
+    if (!configList.isEmpty()) {
```
flippingbits, line 38: The `if` statement might be obsolete.

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
```diff
@@ -0,0 +1,118 @@
...
+      validateConfigList(configList);
+      for (ConfigEntity config : configList) {
+        if (config.getId() != null) {
+          checkValidKind(Kind.STREAM, config.getKind());
```
flippingbits, line 42: I would rename it to better reflect what is being done here, like, `verifyKindOfConfig` or something similar :)

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -161,36 +163,63 @@ public Uni<List<DeploymentEntity>> getDeployments() {
...
+                .transform(
+                    x ->
+                        configList
+                            .onItem()
+                            .transform(
+                                configs -> ConfigUtilities.combineWithDeployment(spec, configs)))
                 .onItem()
                 .transformToUni(
-                    pipelineEntity ->
+                    combinedSpecUni ->
```
flippingbits, line 182: I feel that we should try to keep the endpoints as flat as possible and move the application of configs to deployments and streams out of here.

The `DeploymentEndpoint` should not know about configs. That should be implemented in, for instance, the [K8Deployment](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java) class.

platform-api/src/main/java/io/datacater/core/stream/StreamEndpoint.java
```diff
@@ -57,17 +58,25 @@ public Uni<List<StreamEntity>> getStreams() {
...
+    StreamEntity se = new StreamEntity(stream.name(), stream.spec(), stream.labels());
+    Uni<List<ConfigEntity>> configList =
+        ConfigUtilities.getConfig(ConfigUtilities.getConfigNames(stream.labels()), dsf);
     return dsf.withTransaction(
             (session, transaction) ->
                 session
                     .persist(se)
                     .onItem()
+                    .transformToUni(
+                        a ->
```
flippingbits, line 70: `a` is not helpful here, please use a more descriptive name.

platform-api/src/main/java/io/datacater/core/stream/StreamEndpoint.java
```diff
@@ -57,17 +58,25 @@ public Uni<List<StreamEntity>> getStreams() {
...
+                            Uni.combine()
+                                .all()
+                                .unis(Uni.createFrom().item(a), configList)
+                                .asTuple())
+                    .onItem()
                     .transform(
-                        a -> {
-                          createStreamObject(stream);
-                          return a;
+                        b -> {
```
flippingbits, line 77: `b` is not helpful here, please use a more descriptive name.

flippingbits (changes requested):
Thanks a lot for working on this huge task!

>     * We have a labels field in deployments and streams defined as Map<String, List> to reference more than one item with the same label.

I personally would prefer if we would not use labels in deployments and streams for implicitly referencing a label of a config but introduce something like [labelSelectors](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#label-selectors) from Kubernetes to the specs of deployments and streams. From my point of view, labels and label selectors are two different things and should be treated differently.

>     * When creating a config, we automatically create a label in the Metadata mapping the config name to the label 'app.datacater.io/name'

That works for me 👍 How do we deal with updates?

>     * Config MetaData is saved as a JsonNode in Postgres. To simplify the matching when looking for a referenced config, instead of searching through the metadata, i have mapped it to only look for the referenced name.

Let's keep it simple and fast :) 💯 

I left some more minor feedback, mainly on naming functions. It's probably not the highest priority, but might be helpful.

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
ChrisRousey: How does `extractConfigNames()` or `extractConfigNamesFromLabels()` sound?

> I feel that the function name should indicate that it is searching the labels and not the names of the configs.

The method takes a list of labels, searches them and returns a list of config names extracted from the labels.
My naming logic is "What does the method do -> getConfigNames" with the "extracted from labels" being implicit due to the method parameter.

> It looks like it is always returning one name, so maybe make the name singular?

Lists are possible

flippingbits: My fault - got it! I believe that you can also close this conversation and keep it as it is :)

ChrisRousey:
@flippingbits @HknLof @olis1996 

Thank you guys for your Feedback regarding the resource/config (selector-)label matching.
A quick recap of my takeaways and todos:
- We will move the config referencing in the resources from the `labels` node and add a new optional node `configSelector`
    - This node will be on the same level as the resource spec, not inside the individual resource spec-node
- We will no longer expect the resource `selectorLabels` to contain an array of configs, but rather use the kubernetes way of selector labeling and match key/values with the labels from config.
    - We will no longer "pseudomatch" the referenced configs by there names
    
Feel free to comment if i missed something or my explanation isn't understandable 🙂

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
ChrisRousey: An Empty list can be given if no config is found and the validation method doesn't "break" if the list is empty, since an empty list would be considered a valid config. So this check is still needed

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
ChrisRousey: yes, i do feel the same way. The problem here is that when we retrieve/work with entities with mutiny, it all has to happen inside the worker thread that is returned to the user for execution. If we were to move the config retrieval from here into the K8Deployment class, it would cause a more complicated/distributed nesting making it a lot less readable and harder to work with in my opinion.

platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
flippingbits: I believe it is not needed because you have only two blocks inside this `if` blocks: The validation and the for loop. The validation does not break because it also checks for an empty list and the for loop would just not execute because the list is empty.

Anyway, this is very very low priority. You can also keep it :)

ChrisRousey: you were right 🙂 i rem oved the if-statement

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
ChrisRousey: created issue #174

platform-api/src/main/java/io/datacater/core/stream/StreamEndpoint.java
```diff
@@ -80,20 +84,37 @@ public Uni<Response> createStream(Stream stream) throws JsonProcessingException
...
+                .transformToUni(
+                    streamEntity ->
+                        Uni.combine()
+                            .all()
+                            .unis(Uni.createFrom().item(streamEntity), configList)
+                            .asTuple())
+                .onItem()
+                .transform(
+                    tuple -> {
+                      LOGGER.info(tuple.getItem1());
```
flippingbits, line 104: Can you please add some more information/context to the log message, which helps us to understand what you are logging here? Assume that we see this log message among many other log messages.

platform-api/src/test/java/io/datacater/core/config/ConfigDeploymentTest.java
```diff
@@ -0,0 +1,135 @@
...
+  @Test
+  @Order(3)
+  void postDeployment() {
+    Response response =
+        given()
+            .header("Content-Type", "application/json")
+            .body(deploymentJson.toString())
+            .baseUri(baseURI)
+            .post(deploymentsPath);
+    LOGGER.info("deployment response: ");
```
flippingbits, line 131: Please combine lines 131 and 132 into one line, so we produce one log message.

platform-api/src/test/java/io/datacater/core/config/MultiConfigStreamTest.java
```diff
@@ -0,0 +1,134 @@
...
+  void postStream() throws ExecutionException, InterruptedException {
+    Response response =
+        given()
+            .header("Content-Type", "application/json")
+            .body(streamJson.toString())
+            .baseUri(baseURI)
+            .post(streamsPath);
+
+    LOGGER.info("stream: ");
+    LOGGER.info(response.body().asString());
```
flippingbits, line 108: Please combine lines 107 and 108, so we produce one log message.

flippingbits (changes requested):
LGTM, Thanks a lot for the great work. This is an awesome feature 🏅 

I have some minor feedback on how to use the logger. I prefer to produce single-line log messages, so we can easier investigate them.

flippingbits (approved):

HknLof (approved):

PR #158, opened 2023-02-09
fix(stream-inspect): change message retrieval method

ChrisRousey:
This PR adds a new method to retrieve messages from partitions:
The new method will first get messages from partition1, if the expected amount is not yet reached, the amount is filled with messages from aprtition 2 and so on

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
+++ b/platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
@@ -173,14 +173,14 @@ public class KafkaStreamsAdmin implements StreamService {
    * Inspect (or retrieve) the most recent events of the Topic.
    *
    * @param limit Number of records to retrieve.
-   * @param distributedInspect Mode of operation for the retrieval of records. - true: Distributed
+   * @param distributed Mode of operation for the retrieval of records. - true: Distributed
    *     retrieval. Records are polled evenly across partitions. The return amount can vary
    *     depending on the amount of messages in a partition. - false: Top-down retrieval. Messages
    *     are polled from one partition. If the partition does not contain the amount defined in
    *     `limit`, the next partition is polled and so on.
    * @return a List<StreamMessage> containing the inspected messages form each topic
    */
-  public List<StreamMessage> inspect(Stream stream, long limit, boolean distributedInspect) {
+  public List<StreamMessage> inspect(Stream stream, long limit, boolean distributed) {
     List<StreamMessage> messageList = new ArrayList<>();
     if (!streamExists()) {
       return messageList;
@@ -191,7 +191,7 @@ public class KafkaStreamsAdmin implements StreamService {
     }
     consumer.assign(partitionsList);
 
-    if (distributedInspect) {
+    if (distributed) {
       final long partitionMessageAmount =
           (long) Math.ceil((double) limit / (double) partitionsList.size());
       setPartitionOffsets(partitionsList, partitionMessageAmount);
--- a/platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
+++ b/platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
@@ -173,9 +173,15 @@ public class KafkaStreamsAdmin implements StreamService {
    * Inspect (or retrieve) the most recent events of the Topic.
    *
    * @param limit Number of records to retrieve.
+   * @param distributedInspect Mode of operation for the retrieval of records.
+   *      - true: Distributed retrieval. Records are polled evenly across partitions.
+   *              The return amount can vary depending on the amount of messages in a partition.
+   *      - false: Top-down retrieval. Messages are polled from one partition. If the
+   *               partition does not contain the amount defined in `limit`, the next partition
+   *               is polled and so on.
    * @return a List<StreamMessage> containing the inspected messages form each topic
    */
-  public List<StreamMessage> inspect(Stream stream, long limit) {
+  public List<StreamMessage> inspect(Stream stream, long limit, boolean distributedInspect) {
     List<StreamMessage> messageList = new ArrayList<>();
     if (!streamExists()) {
       return messageList;
--- a/platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
+++ b/platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
@@ -173,12 +173,11 @@ public class KafkaStreamsAdmin implements StreamService {
    * Inspect (or retrieve) the most recent events of the Topic.
    *
    * @param limit Number of records to retrieve.
-   * @param distributedInspect Mode of operation for the retrieval of records.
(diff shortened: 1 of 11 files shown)


ChrisRousey:
FYI i also opened a CR in GitBook for documenting these changes.

platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
```diff
@@ -173,9 +173,14 @@ public Map<String, String> getMetadata() {
    * Inspect (or retrieve) the most recent events of the Topic.
    *
    * @param limit Number of records to retrieve.
+   * @param distributed Mode of operation for the retrieval of records. - true: Distributed
```
flippingbits, line 176: Instead of introducing the boolean flag `distributed`, I suggest introducing a more flexible parameter (or config option) that is defined as a `String` and that can point to an arbitrary number of sampling methods. Maybe call it `sampleApproach` or `sampleMethod`? This would allow us to introduce more sampling approaches later on and make the `inspect` method more flexible in the long term.

flippingbits: Beyond that, to me, `distributed` is a bit confusing as a name for this approach because I immediately think of a distributed algorithm when reading it (maybe it's only me, though?). What about using a different name, like `uniform`? We then need a name for the other approach, too :)

flippingbits (changes requested):

platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
ChrisRousey: thanks for the feedback. i've renamed the variables to uniform and sequenced. feel free to suggest other namings 🙂

platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
```diff
@@ -173,9 +173,14 @@ public Map<String, String> getMetadata() {
    * Inspect (or retrieve) the most recent events of the Topic.
    *
    * @param limit Number of records to retrieve.
+   * @param method Mode of operation for the retrieval of records. - Uniform: Distributed retrieval.
```
flippingbits, line 176: Let's call it `sampleMethod` instead of `method` to align with the naming of the query parameter and be more explicit in general?

platform-api/src/main/java/io/datacater/core/stream/StreamEndpoint.java
```diff
@@ -43,8 +43,10 @@ public Uni<StreamEntity> getStream(@PathParam("uuid") UUID uuid) {
   @GET
   @Path("{uuid}/inspect")
   public Uni<List<StreamMessage>> inspectStream(
-      @PathParam("uuid") UUID uuid, @DefaultValue("100") @QueryParam("limit") Long limit) {
-    return streamsUtil.getStreamMessages(uuid, limit);
+      @PathParam("uuid") UUID uuid,
+      @DefaultValue("100") @QueryParam("limit") Long limit,
+      @DefaultValue("SEQUENCED") @QueryParam("sampleMethod") SampleMethod method) {
```
flippingbits, line 48: I would also rename `method` to `sampleMethod` here, to align with the naming of the query parameter.

platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
flippingbits: Sounds good to me

platform-api/src/test/java/io/datacater/core/stream/StreamInspectStringTest.java
```diff
@@ -55,7 +55,10 @@ void testStringDeserializer()
     lastMessageToWaitOn.toCompletableFuture().get(1000, TimeUnit.MILLISECONDS);
 
     Response response =
-        given().pathParam("uuid", uuid.toString()).queryParam("limit", "3").get("/{uuid}/inspect");
+        given()
+            .pathParam("uuid", uuid.toString())
+            .queryParams("limit", "3", "sampleMethod", SampleMethod.UNIFORM)
+            .get("/{uuid}/inspect");
```
HknLof, line 61: Is this change due to the new query parameter? If so, do we actually test the uniformity of our retrieval, i.e.:

3 Topics hold 30 Items each, if I set limit to 15, I should get 5 from each topic, right?

platform-api/src/test/java/io/datacater/core/stream/StreamInspectMethodsTest.java
```diff
@@ -0,0 +1,110 @@
...
+    Response response =
+        given()
+            .pathParam("uuid", uuid.toString())
+            .queryParams("limit", "10", "sampleMethod", SampleMethod.SEQUENCED)
+            .get("/{uuid}/inspect");
+
+    Assertions.assertEquals(200, response.getStatusCode());
+    Assertions.assertTrue(response.body().asString().contains("\"partition\":0"));
+    Assertions.assertFalse(response.body().asString().contains("\"partition\":1"));
+    Assertions.assertFalse(response.body().asString().contains("\"partition\":2"));
```
HknLof, line 78: Ideally we would count: How many items are from each partition.

platform-api/src/test/java/io/datacater/core/stream/StreamInspectMethodsTest.java
```diff
@@ -0,0 +1,110 @@
...
+    Response response =
+        given()
+            .pathParam("uuid", uuid.toString())
+            .queryParams("limit", "10", "sampleMethod", SampleMethod.UNIFORM)
+            .get("/{uuid}/inspect");
+
+    Assertions.assertEquals(200, response.getStatusCode());
+    Assertions.assertTrue(response.body().asString().contains("\"partition\":0"));
+    Assertions.assertTrue(response.body().asString().contains("\"partition\":1"));
+    Assertions.assertTrue(response.body().asString().contains("\"partition\":2"));
```
HknLof, line 63: Do we return 9 elements or do we have one partition over-represented here?

HknLof (dismissed):

platform-api/src/test/java/io/datacater/core/stream/StreamInspectStringTest.java
ChrisRousey: yes, that's why this test changed. Good idea, i added that test case to [StreamInspectMethodsTests](https://github.com/DataCater/datacater/blob/e5b26109672555a0642d0337db1d9008455b770c/platform-api/src/test/java/io/datacater/core/stream/StreamInspectMethodsTest.java)

platform-api/src/test/java/io/datacater/core/stream/StreamInspectMethodsTest.java
ChrisRousey: We would return 10.
we take the ceiling of 10/3 from each partition then trim the sample size to 10. so we would have:
- Partition 0: 4 records
- Partition 1: 4 records
- Partition 2: 2 records

flippingbits (approved):

Issue #112, opened 2023-01-02
Make Deployment Replicas configurable


ChrisRousey:
Deployment Replicas are currently only configurable at application through the `datacater.deployment.replicas` environment variable.

With this ticket, we aim to make the replica amount confugirable on resource level and to aggregate the health and metrics endpoints to include all deployment replicas

PR #178, opened 2023-03-03
feature: Make deployment replicas configurable on resource level

ChrisRousey:
This PR is to make the deployment replica amount configurable at the resource level, when creating a new deployment.
Ticket reference: #112


ChrisRousey:
The changes of this PR are documentated in a [gitbook PR](https://app.gitbook.com/o/v6uGZYtfBxqXSwxx4vXu/s/FLAe5WA2FuJRguGU5lqo/~/changes/68/docs/api/deployments)

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -341,16 +347,17 @@ private List<String> getDeploymentLogsAsList(UUID deploymentId) {
...
+    String protocol = StaticConfig.EnvironmentVariables.DEPLOYMENT_CONTAINER_PROTOCOL;
+
     String uriReady =
         String.format(
-            "%s://%s:%d%s",
-            StaticConfig.EnvironmentVariables.DEPLOYMENT_CONTAINER_PROTOCOL,
-            clusterIp,
-            StaticConfig.EnvironmentVariables.DEPLOYMENT_CONTAINER_PORT,
-            path);
+            "%s://%s.%s.%s.svc.cluster.local:%d%s", protocol, ip, service, namespace, port, path);
```
flippingbits, line 360: I feel that we could get rid of the service here. From my understanding, we could [directly access the replica via its pod IP](https://kubernetes.io/docs/concepts/services-networking/dns-pod-service/#pods).

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -260,6 +258,50 @@ public Map<String, Object> getDeployment(UUID deploymentId) {
     }
   }
 
+  public String getDeploymentReplicaIp(UUID deploymentId, int replica) {
+    int replicaPosition = replica;
+    if (replicaPosition > 0) {
```
flippingbits, line 263: We should also take care of the case where `replicaPosition <= 0`.

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -260,6 +258,50 @@ public Map<String, Object> getDeployment(UUID deploymentId) {
...
+            .apps()
+            .deployments()
+            .inNamespace(StaticConfig.EnvironmentVariables.NAMESPACE)
+            .withName(deploymentName)
+            .get()
+            .getSpec()
+            .getSelector()
+            .getMatchLabels();
+
+    Pod first;
```
flippingbits, line 279: I feel that we could improve the naming here.

flippingbits (changes requested):
Thanks for the great work :) I left two questions/remarks.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
flippingbits: I think that we no longer need it given that we are directly working with pods. What do you think?

ChrisRousey:
This PR is ready for review @flippingbits @HknLof @olis1996 

I had to remove two tests regarding the logs, since the kubernetes mockserver doesn't actually seem to create real deployments/replicas under the hood. I have created an issue #193 to address this.

platform-api/src/main/java/io/datacater/core/deployment/StaticConfig.java
```diff
@@ -130,7 +124,7 @@ private EnvironmentVariables() {}
     static final String PYTHON_RUNNER_IMAGE_TAG =
         ConfigProvider.getConfig()
             .getOptionalValue("datacater.pythonrunner.image.version", String.class)
-            .orElse("alpha-20221117");
+            .orElse("nightly");
```
flippingbits, line 127: [For other image versions](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/deployment/StaticConfig.java#L95), we point to the latest stable release. Shall we do the same here?

flippingbits (approved):
💯

PR #182, opened 2023-03-09
fix: load image bash postgres config

ChrisRousey:
The current `load-image-k8s.bash` updates the `k8s-manifests/minikube-with-postgres-ns-default.yaml` but actually overwrites/removes the postgres `statefulset` and `service`, which causes the application to crash.

This PR configures postgres when the helm template command is called.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -47,3 +47,4 @@ python-runner/transforms/
 # python
 __pycache__
 .venv
+/k8s-manifests/minikube-with-postgres-image-test.yaml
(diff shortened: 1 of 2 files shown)


flippingbits (dismissed):
Looks good to me. Not sure if there is much need for this script in the future though.

HknLof:
As of now we struggle to configure remote development as we do not find proper documentation on using quarkus remote dev with minikube. This just an easy we to load the change into minikube it is meant for development.

HknLof (approved):

PR #183, opened 2023-03-10
feat(ui/streams): Disable fields that cannot be updated

### LINKED ISSUE #159: Inform users in the UI about fields that cannot be updated
We do not allow updating the fields `name`, `num_partitions`, and `replication_factor` of the Streams resource; they are only writeable when creating a new stream.

We should inform users about that in the UI and disable the corresponding text fields.

flippingbits:
In the UI, disable config fields of streams that do not allow updates.

Fixes #159

### DIFF
--- a/ui/src/containers/streams/EditStream.js
+++ b/ui/src/containers/streams/EditStream.js
@@ -219,8 +219,9 @@ class EditStream extends Component {
                   Name<span className="text-danger ms-1">*</span>
                 </h5>
                 <span className="text-muted fs-7 me-2">
-                  Name of the Apache Kafka topic. If the topic does not yet
-                  exist on the broker, we will automatically create it.
+                  Name of the Apache Kafka topic. You cannot update the field{" "}
+                  <i>name</i> of existing streams. Please create a new stream
+                  instead.
                 </span>
                 <a
                   className="fs-7"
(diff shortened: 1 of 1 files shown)


ui/src/containers/streams/EditStream.js
```diff
@@ -219,8 +219,9 @@ class EditStream extends Component {
                   Name<span className="text-danger ms-1">*</span>
                 </h5>
                 <span className="text-muted fs-7 me-2">
-                  Name of the Apache Kafka topic. If the topic does not yet
-                  exist on the broker, we will automatically create it.
+                  Name of the Apache Kafka topic. You cannot update the field{" "}
```
HknLof, line 222: Why is there a space inserted here? Would it make sens to either insert into `<i>`? Inserting a React construct for having a space in HTML feels unnecessary.

HknLof (changes requested):

ui/src/containers/streams/EditStream.js
flippingbits: In this case, I can blame [prettier](https://prettier.io/) which I ran before committing the changes. 😆

HknLof: Oh wow! :D

HknLof (approved):

PR #185, opened 2023-03-14
feat(ui): Allow resizing of context bar

### LINKED ISSUE #160: Allow to resize the code editor
At the moment, we fix the width of the code editor in the UI.
When working on larger Python functions, it might be convenient if increasing the size of the editor would be possible.

flippingbits:
Allow users to resize the context bar, which is especially helpful when working in the code editor.

To this end, support three sizes: S (500px - default), M (700px), L (900px).

Fixes #160

Screenshot 2023-03-14 at 11 09 05
Screenshot 2023-03-14 at 11 09 11
Screenshot 2023-03-14 at 11 09 14

### DIFF
--- a/ui/src/components/pipelines/pipeline_designer/context_bar/ContextBarSizer.js
+++ b/ui/src/components/pipelines/pipeline_designer/context_bar/ContextBarSizer.js
@@ -19,30 +19,27 @@ class ContextBarSizer extends Component {
 
     return (
       <React.Fragment>
-        <a
-          href="#"
+        <button
           className={sClassNames}
           data-size="S"
           onClick={changeContextBarSizeFunc}
         >
           S
-        </a>
-        <a
-          href="#"
+        </button>
+        <button
           className={mClassNames}
           data-size="M"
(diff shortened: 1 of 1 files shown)


ui/src/components/pipelines/pipeline_designer/context_bar/ContextBarSizer.js
```diff
@@ -0,0 +1,51 @@
...
+      contextBarSize === "L"
+        ? "btn btn-sm btn-dark me-2"
+        : "btn btn-sm btn-outline-dark me-2";
+
+    return (
+      <React.Fragment>
+        <a
+          href="#"
+          className={sClassNames}
+          data-size="S"
```
HknLof, line 25: Is there a reason to not use client-side routing for this?

flippingbits: From my point of view, it does add any value. Let's replace the hyperlinks with `button`s though

HknLof (approved):

PR #189, opened 2023-03-15
feat(UI): Add config functionality

ChrisRousey:
Add the config functionality to the frontend

work in progress

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
+++ b/platform-api/src/main/java/io/datacater/core/config/ConfigUtilities.java
@@ -49,8 +49,12 @@ public class ConfigUtilities {
             .spec()
             .getKafka()
             .putAll(
-                JsonUtilities.toObjectMap(
-                    config.getSpec().get(stream.spec().getKind().name().toLowerCase(Locale.ROOT))));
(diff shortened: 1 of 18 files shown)


flippingbits (commented):
So far, I only looked at the code. It looks good to me 👍. I have two remarks:

* We might want to reduce code duplicates and move the forms for streams and deployments to a component, which can be included by the container components for configs (e.g., `NewConfig.js`) and streams/deployments (e.g., `NewStream.js`). Happy to have a look at this with you together.
* Can you please run `npx prettier --write src` in the `/ui` folder to format the code? We should probably add this as a pre-commit in another PR.

In the next step, I will play with the changes and probably have more feedback :)

HknLof (approved):
Will open issue, once Github works properly for me but just to summarize:

The issue will describe moving away from `<a href="#">` more towards `<Button></Button>`. Especially with the combination of `<a href="#" className="nav-link"></a>` and then actually navigating links not having the class _nav-link_ like `<a href="/configs/new">...</a>` as in Line 64 in ListConfigs.js.

I see it quite often in this PR, but this is an issue with the frontend as a whole, so can be moved out.

PR #191, opened 2023-03-16
PayloadEditor integration for write operations

HknLof:
**Edit-2: April 20th**

The current implementation and state manage follows the approach, that state changes made to the HTML form are synchronized to the payload editor. When changing back form the editor and changes were made, we prompt users to either discard made changes or stay within the editor. If no changes were done, there is no prompt. Generally speaking opening the `Edit as JSON` button can be found at the bottom of the form.

Screenshot 2023-04-20 at 15 13 46

Exceptions to this rule are the *Edit Pipeline* window, *Edit Config*, and *New Config* containers.

For *Edit Pipeline* it would have convoluted our pipeline designer. This container already has lots of contents, hence it *Edit as JSON* is choice before you encounter the *Pipeline Designer*.

Screenshot 2023-04-20 at 15 13 55

*Edit Config* and *New Config* have a slightly special treatment because of the visible tabs. Here, the button *Edit as JSON* is always visible throughout both tabs.

Screenshot 2023-04-20 at 15 14 05

Will update description with screenshots, once editor draft is implemented.

**EDIT-1**: Create a second panel for PUT and POST methods ... @flippingbits, will need your guidance to create the same look and feel as `ApiCall` button. Will implement logic before going into destroying the layout just to fix it again 😄 
Screenshot 2023-03-20 at 15 58 09


HknLof:
@ChrisRousey @flippingbits Grew larger then expected. I did not extract functions for re-use, as this is not being done throughout our UI codebase, [for state management and `this` references, I guess].

ui/src/components/payload_editor/PayloadEditor.js
```diff
@@ -0,0 +1,71 @@
...
+
+import "ace-builds/src-noconflict/mode-json";
+import "ace-builds/src-noconflict/theme-xcode";
+
+export class PayloadEditor extends Component {
+  constructor(props) {
+    super(props);
+    this.state = {
+      didChange: false,
+      newCode: undefined,
```
flippingbits, line 14: I suggest more precise namings here.
For instance, I can't tell what kind of data `newCode` holds.

ui/src/components/payload_editor/PayloadEditor.js
```diff
@@ -0,0 +1,71 @@
...
+      editorCode: ""
+    };
+
+    this.didIntroduceChange = this.didIntroduceChange.bind(this);
+    this.handleChange = this.handleChange.bind(this);
+  }
+
+  didIntroduceChange() {
```
flippingbits, line 22: Again, naming confuses me. This feels like a function that returns a Boolean value, while `this.state.newCode` probably holds something else?

ui/src/containers/configs/NewConfig.js
```diff
@@ -62,6 +63,22 @@ class NewConfig extends Component {
         connectionValue: "",
       },
       configCreated: false,
+      showPayloadEditor: false,
+      payloadEditorChanges: false,
+      editorConfig: {
```
flippingbits, line 68: This seems to be a duplicate from `state.config`. Do we really need it two times?

ui/src/containers/configs/NewConfig.js
```diff
@@ -74,6 +91,12 @@ class NewConfig extends Component {
     this.removeStreamConfig = this.removeStreamConfig.bind(this);
     this.updateConfigSpec = this.updateConfigSpec.bind(this);
     this.updateTab = this.updateTab.bind(this);
+    // payloadEditor specific functions
```
flippingbits, line 94: Thanks for grouping the functions logically. That helps a lot with navigating them 👍

ui/src/containers/configs/NewConfig.js
```diff
@@ -131,6 +154,30 @@ class NewConfig extends Component {
...
+            errorMessage: this.props.config.errorMessage,
+          });
+        } else {
+          this.setState({
+            configCreated: true,
+            errorMessage: "",
+          });
+        }
+      });
+    } catch (syntaxError) {
```
flippingbits, line 173: From my understanding, we are looking for an error that could occur in line 159? If yes, would it make sense to only wrap this line in a `try{}` block?

ui/src/containers/configs/NewConfig.js
```diff
@@ -304,11 +351,64 @@ class NewConfig extends Component {
     });
   }
 
-  render() {
-    if (this.state.configCreated) {
-      return <Redirect to={"/configs/" + this.props.configs.config.uuid} />;
+  toggleForm(event) {
+    event.preventDefault();
+    let toggle = !this.state.showPayloadEditor;
```
flippingbits, line 356: `toggle` does not tell a lot. What about renaming it to `isShowingPayloadEditor`?

ui/src/containers/configs/NewConfig.js
```diff
@@ -304,11 +351,64 @@ class NewConfig extends Component {
...
+        showPayloadEditor: toggle,
+        editorConfig: JSON.stringify(this.state.config, null, 2),
+      });
+    } else if (this.state.payloadEditorChanges && !window.confirm("Going back will reset all edits in the editor!")) {
+      this.setState({
+        showPayloadEditor: true,
+      });
+    } else {
+      this.setState({
+        editorConfig: JSON.stringify(this.state.config, null, 2),
```
flippingbits, line 369: Looks like a duplicate of line 361. Let's merge them by introducing a variable?

flippingbits (changes requested):
Thanks for your work on this great feature! 🥇 

Please see below my feedback:
* I propose removing the file `ui/src/components/payload_editor/PayloadDefaults.js` that seems to be obsolete from my point of view.
* I only left comments in the container component `NewConfig`. The feedback also applies to the other container components, `NewDeployment`, `NewPipeline`, and `NewStream`.

ui/src/components/payload_editor/PayloadEditor.js
HknLof: Removed, since not necessary any longer.

ui/src/containers/configs/NewConfig.js
HknLof: Renamed, thanks for catching this.

ui/src/containers/configs/NewConfig.js
HknLof: Good catch, as we have to treat the editor content as a string only. I replaced it with an empty string. We `JSON.stringify` upon switching to the editor.

HknLof:
Thanks for the feedback, I will put this PR to draft and integrate the PayloadEditor into the `/**/edit`containers.

HknLof:
Integrated the `<PayloadEditor>` into various containers, which perform write operations to our API. Along the way I encountered some issues and housekeeping tasks for our ui implementation, that I would summarize in some tickets.

ui/src/containers/configs/EditConfig.js
```diff
@@ -356,612 +356,703 @@ class EditConfig extends Component {
...
-          </form>
+          </div>
+        </div>
+      </form>
+    );
+  }
+
+  toggleForm(event) {
+    event.preventDefault();
+    let isShowingPayloadEditor = !this.state.showPayloadEditor;
```
flippingbits, line 927: I am a bit confused here. Why do we need to negate `this.state.showPayloadEditor`?

ui/src/containers/pipelines/ShowPipeline.js
```diff
@@ -74,6 +74,12 @@ class ShowPipeline extends Component {
             apiPath={`/pipelines/${pipeline.uuid}`}
             buttons={
               <>
+                <a
+                  href={`/pipelines/${pipeline.uuid}/edit-payload`}
+                  className="btn btn-primary text-white ms-2"
```
flippingbits, line 79: I suggest using `btn-primary` for only one button. What about switching to `btn-secondary` here. What do you think?

ui/src/containers/configs/EditConfig.js
```diff
@@ -356,612 +356,703 @@ class EditConfig extends Component {
...
+      parsedConfig = JSON.parse(this.state.editorConfig, null, 2);
+    } catch (syntaxError) {
+      this.setState({
+        configCreated: false,
+        errorMessage: syntaxError.message,
+      });
+    }
+
+    if (parsedConfig !== undefined) {
+      this.props.updateConfig(parsedConfig).then(() => {
```
flippingbits, line 964: This function [expects two parameter](https://github.com/DataCater/datacater/blob/main/ui/src/actions/configs.js#L108). This way it is not working.

ui/src/containers/streams/EditStream.js
```diff
@@ -198,22 +202,97 @@ class EditStream extends Component {
...
+    } catch (syntaxError) {
+      this.setState({
+        streamUpdated: false,
+        errorMessage: syntaxError.message,
+      });
     }
 
-    const stream = this.state.stream;
+    if (parsedStream !== undefined) {
+      this.props.updateStream(parsedStream).then(() => {
```
flippingbits, line 224: This function [expects two parameter](https://github.com/DataCater/datacater/blob/main/ui/src/actions/streams.js#L108). This way it is not working.

ui/src/containers/deployments/EditDeployment.js
```diff
@@ -149,23 +150,98 @@ class EditDeployment extends Component {
...
+    } catch (syntaxError) {
+      this.setState({
+        deploymentUpdated: false,
+        errorMessage: syntaxError.message,
+      });
     }
 
-    const deployment = this.state.deployment;
+    if (parsedDeployment !== undefined) {
+      this.props.updateDeployment(parsedDeployment).then(() => {
```
flippingbits, line 165: This function expects [two parameters](https://github.com/DataCater/datacater/blob/main/ui/src/actions/deployments.js#L112).

ui/src/containers/pipelines/EditPipelinePayload.js
```diff
@@ -0,0 +1,194 @@
...
+      parsedPipeline = JSON.parse(this.state.editorPipeline, null, 2);
+    } catch (syntaxError) {
+      this.setState({
+        errorMessage: syntaxError.message,
+      });
+    }
+
+    if (this.state.unpersistedChanges && parsedPipeline !== undefined) {
+      return this.props
+        .updatePipeline(this.getPipelineId(), parsedPipeline)
```
flippingbits, line 75: I would suggest redirecting the user to the "ShowPipeline" page after successfully submitting the new pipeline revision, similar to the UX of the other resources.

flippingbits (changes requested):
Thanks for your work. 🚀  I left some feedback.

I couldn't test a couple of components, e.g., EditConfig, EditDeployment, and EditStream, and will have another look after the issues have been fixed.

ui/src/containers/pipelines/ShowPipeline.js
HknLof: I had it as `btn-secondary` it was just confusing to me on why one of the buttons is colored differently as both are taking you to a way of editing the pipeline. But changed it :)

ui/src/containers/configs/EditConfig.js
HknLof: The if conditions below are considering, whether we move to or away from the editor. That way it was possible to avoid some branching and use `isShowingPayloadEditor` also as the next desired state. If you have suggestions on how to improve readability in context of the if statement below, happy to refactor.

ui/src/containers/pipelines/EditPipelinePayload.js
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  render() {
+    const pipeline = this.state.pipeline;
+    if (pipeline == null) {
+      return <></>;
+    }
+
+    if (this.state.pipelineUpdated) {
+      return <Redirect to={`/pipelines/${this.getPipelineId()}`} />;
```
flippingbits, line 92: You need to import the `Redirect` module, otherwise the following error will occur:

```
Uncaught ReferenceError: Redirect is not defined
```

ui/src/containers/pipelines/ShowPipeline.js
```diff
@@ -74,6 +74,12 @@ class ShowPipeline extends Component {
             apiPath={`/pipelines/${pipeline.uuid}`}
             buttons={
               <>
+                <a
+                  href={`/pipelines/${pipeline.uuid}/edit-payload`}
+                  className="btn btn-secondary text-white ms-2"
```
flippingbits, line 79: My apologies for suggesting `btn-secondary` here, it looks a bit weird and hard to read. What about using the same class as for the other buttons, i.e., `btn-light`?

ui/src/containers/configs/NewConfig.js
```diff
@@ -345,598 +440,590 @@ class NewConfig extends Component {
...
-                    href="/configs/new"
-                    className="btn btn-primary text-white mb-4"
-                    onClick={this.handleCreateConfig}
-                  >
-                    Create config
-                  </a>
-                </div>
+              </>
+            )}
+            {![undefined, ""].includes(this.state.errorMessage) && (
```
flippingbits, line 975: For this component, it looks like errors are not shown when using the editor, only when using the form.

flippingbits (changes requested):
Thanks for the new revision 👍 While testing it, I found a few new issues. I feel that you are close to having this PR ready :)

ui/src/containers/pipelines/ShowPipeline.js
HknLof: Assuming, you would like to highlight `Edit in PipelineDesigner`, I used `btn-light btn-secondary` and removed `text-white`. That way, we have a similar look for all buttons apart from `Edit in PipelineDesigner`.

flippingbits: 👍 Looks good, thanks a lot!

PR #192, opened 2023-03-21
fix(api/streams): Return Kafka errors

### LINKED ISSUE #128: Kafka Admin errors regarding false replication factor setting are ignored
If a user sets an incorrect replication factor, there is no error thrown and the topic is added to the datacater frontend. 
We should:
- forward the error to the users
- not create the resource in datacater

flippingbits:
When creating or updating a stream, return errors related to the connection with the associated Apache Kafka broker to the user, e.g., if the replication factor is too high.

Fixes #128


platform-api/src/main/resources/application.yml
```diff
@@ -190,6 +190,7 @@ datacater:
     kafka:
       devservices:
         port: 9092
+        provider: kafka-native
```
ChrisRousey, line 193: I guess this change means we expilictly can't support redpanda brokers. How do we feel about this?

flippingbits: We already took this step for the [pipeline implementation](https://github.com/DataCater/datacater/commit/6b5bc2bb5ecf2408d3eb7eb15bed43acd6d1f1a9). Thus, I don't have any hard feelings about it.

platform-api/src/main/java/io/datacater/core/stream/KafkaStreamsAdmin.java
```diff
@@ -309,7 +329,22 @@ public StreamSpec createStream(StreamSpec spec) {
     if (!streamExists()) {
       NewTopic newTopic = new NewTopic(this.name, this.partitions, this.replication);
       newTopic.configs(spec.getConfig());
-      admin.createTopics(List.of(newTopic));
+      KafkaFuture<Config> createTopicConfigFuture =
+          admin.createTopics(List.of(newTopic)).config(this.name);
```
ChrisRousey, line 333: As far as i can tell, we only call `.config(this.name)` so we receive an object wrapped in `KafkaFuture` to check if the topic could be created.
if we change this line to:
`KafkaFuture<Void> createTopicConfigFuture = admin.createTopics(List.of(newTopic)).all();`
The metadata and configs aren't mapped to the response and we can still use redpanda

flippingbits: That's a great idea. We currently do not check the returned metadata, so happy to implement the suggested change.

PR #198, opened 2023-03-23
feature(ui): add deployment replicas to frontend

ChrisRousey:
This PR introduced the replicas configuration to our frontend newDeployment, editDeployment, newConfig and editConfig resources,

### DIFF
--- a/ui/src/containers/configs/EditConfig.js
+++ b/ui/src/containers/configs/EditConfig.js
@@ -912,6 +912,24 @@ class EditConfig extends Component {
                             }}
                           />
                         </div>
+                        <div className="col-12 mt-2">
+                          <label className="form-label">Replicas</label>
+                          <input
+                            type="text"
+                            className="form-control"
+                            id="replicas"
+                            name="replicas"
+                            onChange={(value) => {
+                              this.handleChange(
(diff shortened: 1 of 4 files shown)


ui/src/containers/configs/EditConfig.js
```diff
@@ -912,6 +916,24 @@ class EditConfig extends Component {
                             }}
                           />
                         </div>
+                        <div className="col-12 mt-2">
+                          <label className="form-label">Replicas</label>
+                          <input
+                            type="text"
```
flippingbits, line 922: Could we use `type="number"` here to indicate the expected data type?

ui/src/containers/configs/NewConfig.js
```diff
@@ -887,6 +891,24 @@ class NewConfig extends Component {
                             }}
                           />
                         </div>
+                        <div className="col-12">
```
flippingbits, line 894: Can you please add a bit of margin to the top by introducing the CSS class `mt-2`?

flippingbits (changes requested):
Thanks a lot for making this available in the UI! :)

Have two small suggestions.

PR #199, opened 2023-03-24
Fix: stream config mapping

ChrisRousey:
This PR is to fix some bugs regarding the config mapping to streams in seperate places throughout the application


platform-api/src/main/java/io/datacater/core/deployment/DeploymentSpec.java
```diff
@@ -28,4 +32,17 @@ public record DeploymentSpec(
...
   }
+
+  public static DeploymentSpec from(DeploymentSpec spec) {
+    Map<String, Object> copiedMap = new HashMap<>();
+    try {
+      JsonNode specNode = DeploymentEntity.serializeMap(spec.deployment());
+      ObjectMapper mapper = new ObjectMapper();
+      copiedMap = mapper.treeToValue(specNode, Map.class);
+    } catch (JsonProcessingException e) {
+      throw new DatacaterException(e.getMessage());
```
flippingbits, line 43: Is the "mapping" to a `DataCaterException` needed for returning a meaningful response to users in the case of an error?

PR #203, opened 2023-03-30
Feature/create deployment status endpoint

olis1996:
An API endpoint has been implemented to get a deployment's status in kubernetes. 
I have a few remaining questions regarding the implementation:

1.  If a user wanted to get a deployment's status by name it would be necessary to call the endpoint something like `deployments/by-name/{deploymentName}/status`, since quarkus would not be able to differentiate it from `deployments/{uuid}/status` without the `by-name` part. Would that implementation be okay or is there a better way that I don't see right now?
2. Is the `io.fabric8.kubernetes.api.model.apps.DeploymentStatus` sufficient as "consistent internal representation across different k8s versions"?
3. Are the tests sufficient?
4. The `testGetDeploymentStatusByValidUuid()` test does not pass and I don't understand why. Inside `testCreateDeployment()` a deployment is created. That deployment's uuid is being used to GET it in the next test `testGetDeployment()`. But for a reason I don't understand the `testGetDeploymentStatusByValidUuid()` returns a 404 although manual tests work just fine

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/deployment/DataCaterDeploymentStatus.java
+++ b/platform-api/src/main/java/io/datacater/core/deployment/DataCaterDeploymentStatus.java
@@ -1,18 +1,12 @@
 package io.datacater.core.deployment;
 
-import com.fasterxml.jackson.annotation.JsonIgnore;
-import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import io.fabric8.kubernetes.api.model.apps.Deployment;
 import io.fabric8.kubernetes.api.model.apps.DeploymentCondition;
 import io.fabric8.kubernetes.api.model.apps.DeploymentStatus;
-import java.util.ArrayList;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
+import java.util.*;
 
 public class DataCaterDeploymentStatus {
-
   @JsonProperty("availableReplicas")
   private Integer availableReplicas;
 
@@ -20,8 +14,7 @@ public class DataCaterDeploymentStatus {
   private Integer collisionCount;
 
   @JsonProperty("conditions")
-  @JsonInclude(JsonInclude.Include.NON_EMPTY)
-  private List<DeploymentCondition> conditions = new ArrayList<DeploymentCondition>();
+  private List<DeploymentCondition> conditions;
 
   @JsonProperty("observedGeneration")
   private Long observedGeneration;
@@ -38,35 +31,35 @@ public class DataCaterDeploymentStatus {
   @JsonProperty("updatedReplicas")
   private Integer updatedReplicas;
 
-  @JsonIgnore private Map<String, Object> additionalProperties = new HashMap<String, Object>();
+  @JsonProperty private Map<String, Object> additionalProperties;
 
   public static DataCaterDeploymentStatus from(Deployment deployment) {
     if (deployment == null) {
       return null;
     }
 
-    // getstatus returns null in tests
     DeploymentStatus deploymentStatus = deployment.getStatus();
 
     if (deploymentStatus == null) {
       return null;
     }
 
-    DataCaterDeploymentStatus self = new DataCaterDeploymentStatus();
-
-    // TODO add default values
-    self.availableReplicas = deploymentStatus.getAvailableReplicas();
-    self.collisionCount = deploymentStatus.getCollisionCount();
-    self.conditions = deploymentStatus.getConditions();
-    self.observedGeneration = deploymentStatus.getObservedGeneration();
-    self.readyReplicas = deploymentStatus.getReadyReplicas();
-    self.replicas = deploymentStatus.getReplicas();
-    self.unavailableReplicas = deploymentStatus.getUnavailableReplicas();
-    self.updatedReplicas = deploymentStatus.getUpdatedReplicas();
-    self.additionalProperties = deploymentStatus.getAdditionalProperties();
-
-    return self;
+    return new DataCaterDeploymentStatus(deploymentStatus);
   }
 
-  private DataCaterDeploymentStatus() {}
+  private DataCaterDeploymentStatus(DeploymentStatus deploymentStatus) {
+    this.availableReplicas = Optional.ofNullable(deploymentStatus.getAvailableReplicas()).orElse(0);
+    this.collisionCount = Optional.ofNullable(deploymentStatus.getCollisionCount()).orElse(0);
+    this.conditions =
+        Optional.ofNullable(deploymentStatus.getConditions()).orElse(new ArrayList<>());
+    this.observedGeneration =
+        Optional.ofNullable(deploymentStatus.getObservedGeneration()).orElse(0L);
+    this.readyReplicas = Optional.ofNullable(deploymentStatus.getReadyReplicas()).orElse(0);
+    this.replicas = Optional.ofNullable(deploymentStatus.getReplicas()).orElse(0);
+    this.unavailableReplicas =
+        Optional.ofNullable(deploymentStatus.getUnavailableReplicas()).orElse(0);
+    this.updatedReplicas = Optional.ofNullable(deploymentStatus.getUpdatedReplicas()).orElse(0);
+    this.additionalProperties =
+        Optional.ofNullable(deploymentStatus.getAdditionalProperties()).orElse(new HashMap<>());
+  }
 }
--- a/platform-api/src/main/java/io/datacater/core/deployment/DataCaterDeploymentStatus.java
+++ b/platform-api/src/main/java/io/datacater/core/deployment/DataCaterDeploymentStatus.java
@@ -13,6 +13,14 @@ public class DataCaterDeploymentStatus {
   @JsonProperty("collisionCount")
   private Integer collisionCount;
 
+  /*
+   DeploymentCondition is json serializable, but ignores additionalProperties. These are properties, which might occur
+   due to changes in the K8s API. Mapping this class to one of our own seems like duplicating the inherent logic of
+   DeploymentCondition.
+
+   Reference:
+   https://github.com/fabric8io/kubernetes-client/blob/master/kubernetes-model-generator/kubernetes-model-apps/src/generated/java/io/fabric8/kubernetes/api/model/apps/DeploymentCondition.java#L77
+  */
   @JsonProperty("conditions")
   private List<DeploymentCondition> conditions;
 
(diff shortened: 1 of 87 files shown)


HknLof:
> 1. If a user wanted to get a deployment's status by name it would be necessary to call the endpoint something like deployments/by-name/{deploymentName}/status, since quarkus would not be able to differentiate it from deployments/{uuid}/status without the by-name part. Would that implementation be okay or is there a better way that I don't see right now?

Go with `UUID's` for now.

> 2. Is the io.fabric8.kubernetes.api.model.apps.DeploymentStatus sufficient as "consistent internal representation across different k8s versions"?

No this is not sufficient. This might change with K8s versions, leading to non-stable API's for our users. Fabric8 upgrades lead to model changes if needed by the supported Kubernetes version.

> 3. Are the tests sufficient?

An interesting test would be testing the payload. Especially, what is returned, if no replicas are created yet?

> 4. The testGetDeploymentStatusByValidUuid() test does not pass and I don't understand why. Inside testCreateDeployment() a deployment is created. That deployment's uuid is being used to GET it in the next test testGetDeployment(). But for a reason I don't understand the testGetDeploymentStatusByValidUuid() returns a 404 although manual tests work just fine.

Not sure what the manual test are from my current reading the following issue is present.
The K8Deployment is never created. https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java#L35
@ChrisRousey might have a quick idea here.

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -255,6 +255,15 @@ public Map<String, Object> getDeployment(UUID deploymentId) {
...
 
+  public Deployment getDeploymentObject(UUID deploymentId) {
+    return client
+        .apps()
+        .deployments()
+        .inNamespace(StaticConfig.EnvironmentVariables.NAMESPACE)
+        .withName(getDeploymentName(deploymentId))
+        .get();
+  }
+
```
ChrisRousey, line 266: by adding this method, we have redundant code in the `getDeployment()` method. Could you edit the getDeployment() to use this method ([K8Deployment.java Lines 244-249](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java#L244-L249))?

platform-api/src/test/java/io/datacater/core/deployment/DatacaterDeploymentEndpointTest.java
```diff
@@ -124,12 +124,36 @@ void testGetDeployment() {
...
   @Test
   @Order(3)
+  void testGetDeploymentStatusByValidUuid() {
+    given()
+        .pathParam("uuid", deploymentId)
+        .baseUri(baseURI)
+        .get(deploymentsPath + "/{uuid}/status")
+        .then()
+        .statusCode(200);
+  }
```
ChrisRousey, line 134: The reason this is failing is because your `[DeploymentEndpoint.getDeploymentStatus()](https://github.com/DataCater/datacater/blob/e109b00e7b5971ae74ac3e99458795231db097d3/platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java#L476)` method is returning null on getStatus(). I beieve this is the case when the deployment is still starting and not yet in  a healthy state. i would suggest using the already implemented [getDeployment()/deploymentToMetaDataMap()](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java#L216-L257) methods, these will return the status if it exists and also some other metadata info, so the user actually gets the information that the deploymet is still starting. Doing that should fix this test

ChrisRousey (commented):
Good progress 👍🏼  I left some comments with opinionated approaches that should help with the failing test

olis1996:
> Not sure what the manual test are from my current reading the following issue is present.

Using the datacater web ui I create a deployment and copy its UUID. In the [SwaggerUI](http://localhost:8080/q/swagger-ui/) I send a GET request to the deployments/{uuid}/status using the copied UUID.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -453,4 +482,18 @@ private void watchLogsRunner(
...
+    Deployment k8Deployment = new K8Deployment(client).getDeploymentObject(uuid);
+
+    if (k8Deployment != null) {
+      return k8Deployment.getStatus();
+    }
+
+    return null;
+  }
+
+   */
```
flippingbits, line 498: Please remove lines 485-498.

platform-api/src/main/java/io/datacater/core/deployment/DataCaterDeploymentStatus.java
```diff
@@ -0,0 +1,72 @@
...
+    // getstatus returns null in tests
+    DeploymentStatus deploymentStatus = deployment.getStatus();
+
+    if (deploymentStatus == null) {
+      return null;
+    }
+
+    DataCaterDeploymentStatus self = new DataCaterDeploymentStatus();
+
+    // TODO add default values
```
flippingbits, line 57: Is the TODO still relevant?

flippingbits (changes requested):
Thanks for working on this issue :)

In general, this already looks good to me. I left two small comments that should be addressed before merging the PR into `main`.

olis1996:
I removed the commented out code + TODO comments, implemented tests for the status endpoint and added default values for the deployment status as follows:

additionalProperties = {},
availableReplicas = 0,
collisionCount = 0,
conditions = [],
observedGeneration = 0,
readyReplicas = 0,
replicas = 0,
unavailableReplicas = 0,
updatedReplicas = 0

Please let me know what you think, especially about the implementation of the default values with optionals. Is that a good approach?

ChrisRousey (changes requested):
Nice work on ths PR, thank you. I do have two small comments though that i didn't see yesterday since the PR was a bit cluttered before you updated from main.

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -133,6 +135,33 @@ public Uni<Response> getMetrics(
...
+              } catch (KubernetesClientException e) {
+                LOGGER.warn(StaticConfig.LoggerMessages.DEPLOYMENT_NOT_FOUND, e);
+                return null;
+              }
+            })
+        .onItem()
+        .ifNull()
+        .failWith(
+            new DeploymentNotFoundException(StaticConfig.LoggerMessages.DEPLOYMENT_NOT_FOUND));
+  }
```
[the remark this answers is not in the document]
olis1996: Thank you for your feedback! You are right, this does not look very clean. I changed this part, so that if a `KubernetesClientException` happens, an unchecked `DeploymentNotFoundException` is thrown. This way I could get rid of the entire `return null` and `.onItem().ifNull().failWith()` part. Please lmk what you think

platform-api/src/main/java/io/datacater/core/deployment/DataCaterDeploymentStatus.java
```diff
@@ -0,0 +1,65 @@
...
+
+public class DataCaterDeploymentStatus {
+  @JsonProperty("availableReplicas")
+  private Integer availableReplicas;
+
+  @JsonProperty("collisionCount")
+  private Integer collisionCount;
+
+  @JsonProperty("conditions")
+  private List<DeploymentCondition> conditions;
```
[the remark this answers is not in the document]
HknLof: I would prefer to have an internal representation :) Could also be just a simple class / enum / record type holding a single string field.

HknLof:
Thanks, the PR looks good. The last bit would be to have an DataCater internal representation for `DeploymentCondition`. I love the exhaustive tests looking into the response

platform-api/src/main/java/io/datacater/core/deployment/K8Deployment.java
```diff
@@ -256,6 +250,19 @@ public Map<String, Object> getDeployment(UUID deploymentId) {
...
+  public Deployment getDeploymentObject(UUID deploymentId) throws KubernetesClientException {
+    try {
+      return client
+          .apps()
+          .deployments()
+          .inNamespace(StaticConfig.EnvironmentVariables.NAMESPACE)
+          .withName(getDeploymentName(deploymentId))
+          .get();
+    } catch (KubernetesClientException e) {
+      throw new DeploymentNotFoundException(StaticConfig.LoggerMessages.DEPLOYMENT_NOT_FOUND);
```
flippingbits, line 262: There might be other reasons for a `KubernetesClientException` besides the non-existence of the requested deployment. With the current implementation, we lose this information.

Could we either log the message of the `KubernetesClientException` before re-throwing it as a `DeploymentNotFoundException` or pass the message to the `DeploymentNotFoundException`?

flippingbits (commented):
Thanks a lot for the great work! I like it a lot. I left one comment, which you might want to address before merging the PR.

platform-api/src/main/java/io/datacater/core/deployment/DataCaterDeploymentStatus.java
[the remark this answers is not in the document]
HknLof: Sorry for the last sentence and the confusion. We will deliver `List<DeploymentCondition> conditions;`, but ensure to have commented the reason for relying on an external model.

HknLof (dismissed):
Thanks for bringing this over the finish line!

ChrisRousey (approved):

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
```diff
@@ -133,6 +134,24 @@ public Uni<Response> getMetrics(
                 }));
   }
 
+  @GET
+  @Path("{uuid}/status")
+  @Produces(MediaType.APPLICATION_JSON)
```
flippingbits, line 139: Do we need this annotation? Looks like we already defined it on the class level.

flippingbits (approved):

platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java
olis1996: You're right, the annotation is not needed

flippingbits (approved):
Thanks for resolving the minor issue, let's merge this into main! :)

PR #204, opened 2023-03-31
feat(ui/deployments): Allow selecting a deployment replica

flippingbits:
We support running deployments with multiple replicas.

This commit adds support for selecting a specific replica when viewing a deployment.

The UI shows the health/metrics/logs of the selected replica.

By default, replica 1 is selected.

## Impact on existing behavior

This PR does not break any existing behavior. At the moment, the UI does not specify the replica when calling the backend, thus it always shows information from the [default replica](https://github.com/DataCater/datacater/blob/main/platform-api/src/main/java/io/datacater/core/deployment/DeploymentEndpoint.java#L69) 1.

## How it looks in action

Choosing a replica when viewing a deployment:
Screenshot 2023-03-31 at 10 27 12

Choosing a replica when viewing the logs of a deployment:
Screenshot 2023-03-31 at 10 27 36

### DIFF
--- a/ui/src/components/deployments/ReplicaSelect.js
+++ b/ui/src/components/deployments/ReplicaSelect.js
@@ -10,7 +10,11 @@ class ReplicaSelect extends Component {
       deployment.spec === undefined ||
       deployment.spec.replicas < 1
     ) {
-      return <></>;
+      return (
+        <button className="btn btn-light ms-2" disabled={true}>
+          No replicas available
+        </button>
+      );
     }
 
     const replicas =
--- a/ui/src/components/deployments/ReplicaSelect.js
+++ b/ui/src/components/deployments/ReplicaSelect.js
@@ -1,0 +1,53 @@
+import React, { Component } from "react";
+import Select from "react-select";
+
+class ReplicaSelect extends Component {
+  render() {
+    const { currentReplica, deployment, updateCurrentReplicaFunc } = this.props;
+
+    if (
+      deployment === undefined ||
+      deployment.spec === undefined ||
+      deployment.spec.replicas < 1
+    ) {
+      return <></>;
+    }
+
+    const replicas =
+      deployment.spec === undefined || isNaN(parseInt(deployment.spec.replicas))
+        ? 1 // By default, deployments have 1 replica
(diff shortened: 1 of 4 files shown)


ui/src/components/deployments/ReplicaSelect.js
```diff
@@ -0,0 +1,53 @@
...
+class ReplicaSelect extends Component {
+  render() {
+    const { currentReplica, deployment, updateCurrentReplicaFunc } = this.props;
+
+    if (
+      deployment === undefined ||
+      deployment.spec === undefined ||
+      deployment.spec.replicas < 1
+    ) {
+      return <></>;
```
HknLof, line 13: Shouldn't we display the button and grey it out? Maybe with `No Available Replicas`?

ui/src/containers/deployments/DeploymentLogs.js
```diff
@@ -117,7 +130,14 @@ class DeploymentLogs extends Component {
           />
           <Header
             apiDocs="https://docs.datacater.io/docs/api/deployments/"
-            apiPath={`/deployments/${deployment.uuid}/logs`}
+            apiPath={`/deployments/${deployment.uuid}/logs?replica=${this.state.currentReplica}`}
```
HknLof, line 133: I often feel like, these should be routes in react :D So, that view selection happens automatically instead of us having to re-create state management.

ui/src/components/deployments/ReplicaSelect.js
flippingbits: Thanks, good point! We now render a button with `disabled={true}` if `spec.replicas` equal 0.

Screenshot 2023-03-31 at 11 23 42

ui/src/containers/deployments/DeploymentLogs.js
flippingbits: I agree in general. It's not always possible to turn a frontend URI (or React route) into an API path though.

Shall we approach this in a separate PR/issue?

HknLof: Yes, let us approach in a separate PR.

HknLof (approved):

Issue #177, opened 2023-03-02
Support YAML in API endpoints


flippingbits:
At the moment, our API endpoints speak only JSON.

We should introduce the support for YAML because a lot of our users prefer YAML over JSON for defining, for instance, pipeline specifications.

We should not break the current behavior and stick to JSON as the default data format. However, users should be able to use YAML as the data format for their API calls by setting the request headers `Content-Type` and/or `Accept` to `application/x-yaml`.

HknLof:
As of Quarkus 2.16.5.Final it should be possible to use the `jackson-jaxrs-yaml-provider`, which automatically uses the right Serde format.

https://github.com/quarkusio/quarkus/issues/31818

flippingbits:
Fixed by #210

PR #209, opened 2023-04-17
build: update Quarkus to most recent version

ChrisRousey:
This PR is to upgrade quarkus to the most recent version. Ticket #177 needs at least quarkus 2.16.5.FINAL


flippingbits (approved):
Thanks for keeping the dependency up-to-date :)

PR #210, opened 2023-04-18
Feat: support yaml payloads

ChrisRousey:
This PR allow the usage of yaml payloads when creating and updating resources in datacater.


flippingbits (commented):
Thank you for this great work. This will make a lot of developers happy :)

Would it make sense to also add a test case that uses different MIME types for the request and the response? For instance, one could set `Content-Type: application/json` and `Accepts: application/x-yaml`.

flippingbits (approved):
Thank you for adding the tests! :)

PR #213, opened 2023-04-27
fix: change timeout in tests

ChrisRousey:
This PR is to fix some tests that *sometimes* fail.

Before:
- We would send records to kafka (redpanda)
- We would wait unteil the last record has been acked to proceed with th tests
- If the last record took longer than our timeout, the test would fail

Now:
- We have remove the timeout to wait for the last record
- We now use `thread.sleep()` to wait so that most, is not all, records have been sent.
- This should solve the problem of the sporadically failing tests.

### DIFF
--- a/platform-api/src/test/java/io/datacater/core/stream/StreamInspectAvroTest.java
+++ b/platform-api/src/test/java/io/datacater/core/stream/StreamInspectAvroTest.java
@@ -14,6 +14,7 @@ import java.io.IOException;
 import java.net.URL;
 import java.util.Properties;
 import java.util.UUID;
+import java.util.concurrent.*;
 import javax.inject.Inject;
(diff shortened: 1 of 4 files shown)


ChrisRousey:
@HknLof 
The tests used to test the validity of the last written record, so we would wait 1-2 seconds to ensure that it was written before checking. 
We aren't checking the validity of underlying records anymore, but only checking that records could be retrieved. To ensure that any records have been written, i added `thread.sleep()` to the tests. This seems kind of dirty, do you have any other ideas? @flippingbits had a valid objection that these tests could also fail in the future if the latency goes up during tests.

flippingbits (approved):
Thanks for incorporating our feedback and avoiding the usage of `Thread.sleep()`.

PR #216, opened 2023-05-22
fix: re-add security annotations for streams

ChrisRousey:
with PR #210 support yaml payloads, the security annotations were removed for testing purposes and never re-added


flippingbits (approved):
Thanks for catching and fixing the bug!

PR #218, opened 2023-05-23
Refactor: stream, deployment, config and pipeline

ChrisRousey:
This PR refactors all the top level resources in datacater.

Refactors include:
- Moving all non-http methods from the endpoint classes to respective utility classes
  - pipeline resources in pipeline utility, stream resources in stream utility, etc.
- adding static config classes where not already existant
- removing "magic variables" to the static config classes
- refining some logging messages
- adding logging statements where I saw fit 
- added a logger utility class for generic logger methods

Please feel free to make more suggestions.

### DIFF
--- a/platform-api/src/main/java/io/datacater/core/stream/StreamUtilities.java
+++ b/platform-api/src/main/java/io/datacater/core/stream/StreamUtilities.java
@@ -1,36 +1,32 @@
 package io.datacater.core.stream;
 
 import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.JsonNode;
+import com.fasterxml.jackson.databind.ObjectMapper;
 import io.datacater.core.authentication.DataCaterSessionFactory;
 import io.datacater.core.config.ConfigEntity;
 import io.datacater.core.config.ConfigUtilities;
+import io.datacater.core.deployment.DeploymentSpec;
+import io.datacater.core.exceptions.CreateDeploymentException;
 import io.datacater.core.exceptions.DatacaterException;
 import io.datacater.core.exceptions.DeleteStreamException;
+import io.datacater.core.pipeline.PipelineEntity;
 import io.datacater.core.utilities.LoggerUtilities;
 import io.smallrye.mutiny.Uni;
 import io.smallrye.mutiny.unchecked.Unchecked;
 import java.util.List;
+import java.util.Map;
 import java.util.UUID;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.TimeoutException;
 import javax.enterprise.context.ApplicationScoped;
 import javax.inject.Inject;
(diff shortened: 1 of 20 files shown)


flippingbits (commented):
Thanks for your effort and time. We get easily drown in working on new features, while maintaining our code base and taking care of refactoring existing code is equally important.

Some of these utilities contain functionality that is also needed in DataCater EE. I am wondering whether it is feasible and would make sense to _outsource_ such functionality into an own _utility_ project, similar to what we did with the [SerDe](https://github.com/DataCater/datacater/tree/main/serde) classes, such that we could include the shared functionality in both the open core and ee and effectively get rid of duplicated code.

ChrisRousey:
> Some of these utilities contain functionality that is also needed in DataCater EE. I am wondering whether it is feasible and would make sense to _outsource_ such functionality into an own _utility_ project, similar to what we did with the [SerDe](https://github.com/DataCater/datacater/tree/main/serde) classes, such that we could include the shared functionality in both the open core and ee and effectively get rid of duplicated code.

@flippingbits do you have any examples of which utility methods could be shared off the to of you head so i can look further into this? Otherwise i'll have a look myself.
In general I think this is a good idea. I'd have to test how smoothly we could migrate mutiny worker methods like `StreamUtilities.getStreamFromDeployment()` since they are entity dependant. We might also lose some sonarqube analysis metrics for the utilities, stuff like test coverage might not work correctly since, my guess is, we'd have to implement the tests in the main repo?
But those are just guesses i'd have to confirm by testing this first.

HknLof:
Just to jump in with another comment  :D. From a Java Quarkus perspective, I do believe, that these would have to be externalized into Quarkus extensions or published as JARs.

flippingbits:
> @flippingbits do you have any examples of which utility methods could be shared off the to of you head so i can look further into this? Otherwise i'll have a look myself.

Yes, `ConfigUtilities` or `StreamUtilities` are examples of functionality that might be needed in EE. However, I would suggest to wait with outsourcing these classes for now, since we currently discuss other ways for a more efficient integration of the open core and EE. If it stays relevant, we could address it in a separate PR.

platform-api/src/main/java/io/datacater/core/stream/StreamUtilities.java
```diff
@@ -0,0 +1,182 @@
...
+import javax.inject.Inject;
+import org.jboss.logging.Logger;
+
+@ApplicationScoped
+public class StreamUtilities {
+  private static final Logger LOGGER = Logger.getLogger(StreamUtilities.class);
+  @Inject DataCaterSessionFactory dsf;
+
+  public Uni<List<StreamMessage>> getStreamMessages(UUID uuid) {
+    return getStreamMessages(uuid, 100L, SampleMethod.SEQUENCED);
```
flippingbits, line 33: Can we move the `100L` into a static config?

flippingbits (approved):

PR #220, opened 2023-05-24
fix(python-runner): Update requests to >= 2.31.0

flippingbits:
Dependabot [reported](https://github.com/DataCater/datacater/security/dependabot/15) an unintended leak of the `Proxy-Authorization` header in the version of the `requests` library that we are using (all versions < 2.31.0 are affected).

This commit updates the `python_runner/requirements.txt` and requires a version of `requests` where this issue has been fixed (>= 2.31.0).


HknLof:
Is there a way to define the python version in requirements or somewhere else? I ran into the issue, that some packages were not available.

Issue #190, opened 2023-03-16
Define and refine logging principles


ChrisRousey:
As of now, there is no process or guideline on when, where, and how to go about logging in DataCater. With the current state, it can be hard to find useful logging information when working with the product in end-to-end scenarios or when debugging a problem.

In order to make the user and developer experience more robust, we should:
- Outline logging guidelines for current and future contributers inside the `CONTRIBUTING.md` file
    - What information should be logged?
    - When should information be logged?
    - How should information be logged?
- Apply these guidelines to the current codebase
    - Frontend
    - platform-api
    - pipeline

PR #224, opened 2023-06-07
docs: create logging guidelines

ChrisRousey:
This PR aims to create a common understanding and guideline on when and where to log application information.
This should provide a better user and developer experience for anyone using DataCater.

Please give me feedback on the guidelines i have added and feel free to suggest changes or anything that should be added to the guidelines.

refs: #190

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -214,10 +214,25 @@ some example of when to NOT log information:
 - A standard object has been created (such as a pipeline object)
 - A standard application process has begun (such as `starting validation process`)
 
+#### Automatic Logging
+Our framework takes care of some standard logging itself. These being:
+- At startup, things regarding configurations and services are logged
+- Kafka Admin Client and Consumer write various logs
+  - When creating topics, configuration warnings are logged
+  - When creating or polling topics, errors are automatically logged
+- Long-running threads are logged
+- Exceptions are logged
+
+When considering when and what to log, always take the extensions you are working with and their capabilities into consideration.
+Logging potential errors and warnings is always a good idea, but when these things are logged multiple times
+it can cause confusion for administrators, users and developers and potentially waste time when analysing the system behaviour.
+
+
 ### What information should be logged
 #### Exception Log Messages
-We should aim to log exception messages in order to give users and admins context and
-more information on what potentially went wrong.
+We should aim to log all exception messages in order to give users and admins context and
+more information on what potentially went wrong. Avoid silencing exceptions by, for example, catching
+an error and not throwing the exception or at least logging that something went wrong.
 The following questions should be answered with these log messages.
 - Why was the exception thrown by DataCater?
 - What parameter led to the Exception?
@@ -295,7 +310,7 @@ WARN:
 ```Java
 if(true){...}
 else {
-    LOGGER.warn("'I don't repsond to threats, I make them' - The Joker");
+    LOGGER.warn("'I don't respond to threats, I make them' - The Joker");
 }
 ```
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -248,10 +248,11 @@ some example of when to NOT log information:
 
 ### How information should be logged.
 #### Log in JSON Format
-We should aim to always log messages in JSON format. JSON format is not only easily user readable
+We should aim to always log messages in JSON format for production environments. JSON format is not only easily user readable
 it also allows for seamless integrations with external log management tools for extra flexibility
 and a higher observability.
-The logging framework we use at DataCater automatically appends the logger messages into JSON format.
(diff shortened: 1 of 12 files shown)


CONTRIBUTING.md
```diff
@@ -175,3 +175,90 @@ For testing and development you can find self-signed certificates under
...
+- how information should be logged.
+
+### What information should be logged.
+#### Unknown-Error Log Message
+These types of messages are things such as an unexpected exception was thrown, which should
+provider relative information and context on the exception.
+The following questions should be answered with these log messages.
+- What error happened
+- what data was used to cause this error
+- when did the error happen
```
HknLof, line 195: The list does not have to be exhaustive, but should start clear and concise. Hence, I would start with:

```
1. Why was the exception thrown by DataCater?
2. What parameter led to the Exception? In e.g. `URIInvalidException`, because of invalid String for an URI, then the String should be printed out. [No user information is allowed to be printed out]
```
That would be my suggestion. Feel free to suggest another list. Let us avoid `- ...` to avoid ambiguity.

CONTRIBUTING.md
```diff
@@ -175,3 +175,90 @@ For testing and development you can find self-signed certificates under
...
+- ...
+
+
+#### Informational Log Message
+These types of log messages are used to give application administrators contexts to user errors,
```
HknLof, line 200: Should we let JAX-RS log all incoming requests? If so, how do we configure it to comply with our requirements?

CONTRIBUTING.md
```diff
@@ -175,3 +175,90 @@ For testing and development you can find self-signed certificates under
...
+users errors can be for example, a stream could not be created due to incorrect user data.
+The following questions should be answered with these log messages.
+- when did the error occur
+- what data was used to cause the error
+- what api path did the user call
+- ...
+
+#### Exclude sensitive information
```
HknLof, line 208: Solid recommendation! :+1:

CONTRIBUTING.md
```diff
@@ -175,3 +175,90 @@ For testing and development you can find self-signed certificates under
...
+Be careful when working on parts of the application that could internationally contain sensitive data.
+In these cases, it could be useful to not include the data payload when logging.
+In some cases, an object id could be provided for more context, such as a Tenant/UserID instead
+of a username but only where the id belongs to a persisted object that an administrator can access.
+
+#### Avoid adding too much Context
```
HknLof, line 214: We should define the minimum necessary amount of information for context. In e.g. _"As a DataCater developer I will know where to look and debug in the codebase by reading the error message in the logs."_

CONTRIBUTING.md
```diff
@@ -175,3 +175,90 @@ For testing and development you can find self-signed certificates under
...
+Having too much information in a log message is a thing. Image a deploymentID being logged for
+an object before it has been saved to the database. The error causes it to not be persisted, but the
+admin assumes it should be in the DB. This would waste time in searching for the object
+and potentially cause the assumption of an error that doesn't actually exist.
+
+### When information should be logged.
```
HknLof, line 220: I like this section a lot. This should go at the top of this document. Further, we should do housekeeping by defining, what is getting automatically logged and where I as a developer need to add logging statements myself.

CONTRIBUTING.md
```diff
@@ -175,3 +175,90 @@ For testing and development you can find self-signed certificates under
...
+
+### How information should be logged.
+#### Log in JSON Format
+We should aim to always log messages in JSON format. JSON format is not only easily user readable
+it also allows for seamless integrations with external log management tools for extra flexibility
+and a higher observability.
+The logging framework we use at DataCater automatically appends the logger messages into JSON format.
+
+
+#### Use the correct logging level:
```
HknLof, line 257: This great. Maybe we can enhance with examples, like 
``` java
try {...} 
catch (SomeRuntimeException sre) { 
  LOGGER.debug("Exception occured, although I am a genius developer.");
  throw sre;
}
```

Maybe give a small example for each level. For info this might somehting like:

```java
public void startup(@Observes StartupEvent event) {
  LOGGER.info("Loading xyz during startup. Stay tuned ...");
}
```

HknLof (dismissed):

HknLof:
Thanks for adding this section, looks like a good first draft. I left some comments feel free to apply to the document the comments you feel are appropriate. :)

CONTRIBUTING.md
ChrisRousey: I personally don't like logging info level logs by default. At least not info logs for core parts of the application. I feel like it clutters more than add context.

My opinion towards this isn't very strong though, so if you consider it helpful i'll gladly add the functionality.

HknLof (approved):

PR #226, opened 2023-06-12
fix: fix injection bug introduced after refactoring pipelines

ChrisRousey:
After refactoring our pipelines resource, we were injecting the `PythonRunnerPool` inside the utilities class and trying to access it in the `PipelineEndpoint` through the injected `PipelineUtilities`.

This Fix removes this detour and injects the `PythonRunnerPool` directly in the `PipelineEndpoint` class.


flippingbits (approved):
Thanks for fixing the issue so fast!

PR #227, opened 2023-06-13
show loading spinner (/throbber) after clicking a button while request is processing

olis1996:
A spinning circle has been implemented which is displayed on the "create stream" button during request processing. There is one thing about my implementation which I'm not happy about: the spinner's svg file can't be placed inside the `public/images` folder, because the framework (?) only allows resources which are located in the `src` folder. 

I added a 2 second delay after clicking the button, so the animation is visible for a longer period.  This is supposed to make it easier to give feedback and I will remove that delay before completing the PR.

What do you think about my approach? I'm really looking forward to your feedback.


flippingbits (changes requested):
Thanks for your efforts! 

A few remarks and change requests from my side:

* Instead of maintaining the status of the create stream request in the container component, please extend the [streams reducer](https://github.com/DataCater/datacater/blob/main/ui/src/reducers/streams.js#L4) with something similar to the boolean `fetchingStreams` variable, e.g., you could introduce `creatingStream`. Then, you can access it inside the container component via `this.props.streams.creatingStream`.
* We use Bootstrap in the frontend, which provides a [Spinner component](https://getbootstrap.com/docs/5.3/components/spinners/). I suggest using that one instead of introducing a custom spinner.
* Please introduce the same functionality to the [EditStream](https://github.com/DataCater/datacater/blob/main/ui/src/containers/streams/EditStream.js) container component.

olis1996:
Thank you very much for your feedback. I adapted my implementation according to your requests. There is two minor things left, which bother me:
1) The button changes its width depending on which content is displayed (text or spinner). I tried a fixed size of 150px, which worked, but I don't think that is the desired approach. 
2) The spinner is not vertically centered and I dont know how to change that. (I suppose the text is not centered either, but it doesn't bother in that case)

+ It is not possible to disable the "Update stream" anchor element while waiting for the response. I guess it is not easily changed to a button element, right?

olis1996:
Create + Edit Pipeline + Deployment buttons now have a loading spinner, too

HknLof (dismissed):
Thanks for the job here!

flippingbits (approved):
Thanks for the refinements, folks!

HknLof (approved):

PR #230, opened 2023-06-22
feat(api/deployments): Forward health response status

flippingbits:
The `/api/v1/deployments/{uuid}/health` endpoint returns the health status of the underlying Kubernetes Deployment.

So far, the endpoint returns the health information in the response body and always uses the HTTP status code 200 (OK) for the response. This is suboptimal and does not allow for a straightforward integration of the endpoint with observability and monitoring products, which often inspect the response status code (and not the response body) for evaluating whether a service is healthy.

This commit changes the behavior of the endpoint as follows:

* If the Kubernetes Deployment is healthy, it returns the HTTP status OK (200).
* If the Kubernetes Deployment is not healthy, it returns the HTTP status INTERNAL SERVER ERROR (500).


ChrisRousey (approved):
Works fine and looks good to me 🙂

flippingbits:
Thanks for taking a look at it! :)

